PR #2, opened 2021-03-15
Official Robocup world

thomas16441:
This PR adds the official world files supplied by Webots


ysims (commented):
Might be better to separate into two PRs for the servos?
Should update the README too.

ysims:
When I run it there are a load of warnings about not finding files in the right places. 
You've also left your own controller for the robot linked, which doesn't exist in this PR.

ysims:
I'm good for this to be merged now, probably check that the changes I put in work and that there are no warnings or anything anymore @thomas16441

ysims (approved):

PR #3, opened 2021-03-15
PID control gains for motors

TheCoolKidsTable:
This PR adds the PID gains sourced from the Simscape simulation


ysims (approved):
Bit hard to tell if it's doing anything, but nothing seems to break. You mentioned getting it to stand - would be good to get it to do that and stop falling over.

PR #7, opened 2021-03-23
Removes copying the controller executable to the project root

ysims:


Bidski (changes requested):
Removing this isnt correct. Webots [requires](https://cyberbotics.com/doc/guide/programming#what-does-this-mean-could-not-find-controller---loading-void-controller-instead--) the binary to be in a specific location.

The better option is to actually generate the file in the correct location rather than copying it. Something like this
```
set_target_properties(nugus_controller PROPERTIES RUNTIME_OUTPUT_DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}"
                                       OUTPUT_NAME nugus_controller)
```

Otherwise, just update the copy command to something like this
```
add_custom_command(TARGET nugus_controller POST_BUILD COMMAND ${CMAKE_COMMAND} -E
  copy "$<TARGET_FILE:nugus_controller>" "${CMAKE_CURRENT_SOURCE_DIR}/nugus_controller"
)
```

controllers/NUgus_Controller/CMakeLists.txt
```diff
@@ -10,8 +10,5 @@ add_executable(nugus_controller ${controller_sources})
 # Link against the webots target (this will also add any necessary include directories to our target)
 target_link_libraries(nugus_controller PRIVATE webots::webots)
 
-# Copy the built executable to the project root directory
-add_custom_command(TARGET nugus_controller POST_BUILD COMMAND ${CMAKE_COMMAND} -E
-  copy ${CMAKE_CURRENT_BINARY_DIR}/nugus_controller ${CMAKE_SOURCE_DIR}
-)
-
+set_target_properties(nugus_controller PROPERTIES RUNTIME_OUTPUT_DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}"
+                                       OUTPUT_NAME nugus_controller)
```
Bidski, line 14: Until we do the rename. Also, update the gitignore file 
```suggestion
set_target_properties(nugus_controller PROPERTIES RUNTIME_OUTPUT_DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}"
                                       OUTPUT_NAME NUgus_Controller)
```

Bidski (changes requested):

Bidski (approved):
I think this should be ok

PR #1, opened 2021-03-11
Networking for webots

CMurtagh-LGTM:

PR #8, opened 2021-03-24
Controller to teleport robot around playing field to gather image training data

Lachourt:


controllers/teleport_controller/teleport_controller.cpp
```diff
@@ -0,0 +1,202 @@
...
+// Description:   Method to teleport robot around the playing field for the purposes of
+//                    gathering training data
+// Author:        Lachlan Court
+// Modifications:
+
+
+#include <RobotisOp2GaitManager.hpp>
+#include <RobotisOp2MotionManager.hpp>
+#include <webots/Supervisor.hpp>
+//#include <webots/PositionSensor.hpp>
```
tayloryoung6396, line 12: Probs remove commented out code

tayloryoung6396 (changes requested):

controllers/teleport_controller/teleport_controller.cpp
Lachourt: Done

wongjoel:
Is the binary file [controllers/teleport_controller/teleport_controller](https://github.com/NUbots/Webots/blob/35131dcde9d3bdc340635091390617fbb04912a1/controllers/teleport_controller/teleport_controller) meant to be part of the PR? It looks like the compiled binary of the cpp file.

Lachourt:
> Is the binary file [controllers/teleport_controller/teleport_controller](https://github.com/NUbots/Webots/blob/35131dcde9d3bdc340635091390617fbb04912a1/controllers/teleport_controller/teleport_controller) meant to be part of the PR? It looks like the compiled binary of the cpp file.

Lachourt:
I think you're right, I didn't think about that when I made the commit but I can remove it if you like

wongjoel:
> I think you're right, I didn't think about that when I made the commit but I can remove it if you like
> Also oops didn't mean to close the PR still figuring out how this works

All good, I'm still trying to work out how to make comments on particular lines/files. If the binary file doesn't have a specific purpose, it's probably best to remove it.

controllers/teleport_controller/teleport_controller.cpp
```diff
@@ -0,0 +1,138 @@
...
+static const int rotationsSize = 4;
+static const double rotations[rotationsSize][4] = {{1, 0, 0, 1.57081},
+                                                  {0.577353, 0.577345, 0.577353, 2.0944},
+                                                  {1.69526e-09, 0.707103, 0.707111, 3.14159},
+                                                  {-0.577352, 0.577347, 0.577352, -2.09441}};
+
+*/
+
+// Rotations matrix for 45 degree rotations
+static const int rotationsSize = 8;
```
KipHamiltons, line 30: numeric constants would normally use screaming snake case, but we probably won't keep this one (`ROTATIONS_SIZE` would be preferred if we did keep it). You do this so that the developer knows when they attempt to write to a variable in that case that they are doing something wrong.

ysims (commented):
- Edit the name of the PR so it isn't the branch name. Make it descriptive as to what the PR does.
- If the PR name doesn't describe everything, add info in the description of the PR
- Change from using a manual makefile to using cmake, for consistency. You can probably copy the cmakelists.txt from the other controller, just change from nugus_controller to teleport_controller. Ideally, we'd have a script that automatically generates the cmake stuff for new controllers but that is beyond the scope of this PR.

controllers/teleport_controller/teleport_controller.cpp
```diff
@@ -0,0 +1,138 @@
...
+static const double rotations[rotationsSize][4] = {{1, 0, 0, 1.57081},
+                                                  {0.577353, 0.577345, 0.577353, 2.0944},
+                                                  {1.69526e-09, 0.707103, 0.707111, 3.14159},
+                                                  {-0.577352, 0.577347, 0.577352, -2.09441}};
+
+*/
+
+// Rotations matrix for 45 degree rotations
+static const int rotationsSize = 8;                                     
+static const double rotations[rotationsSize][4] = {{1, 0, 0, 1.57081},
```
KipHamiltons, line 31: There are a few things I would change about this approach, and other things I would tell you for future reference. In no particular order:
- these constants are known at compile time, so we could (if we were going to keep them here (we're probably not going to)) make them `constexpr`'s instead of just `const`. `constexpr` is an even nicer version of const, but it has to be knowable at compile time.
- We don't usually use arrays like this. `std::array` is nicer, if we want to use arrays, but that's quite rare. More often we would use `std::vector`, because dynamic sizing is nice (these are basically C++'s version of java's arraylists). In this case, in the main codebase we would probably use a `std::array` of one of eigen's vector types (our linear algebra library of choice), since the number of rotations is constant. I am not sure if this repo will have eigen any time soon though, so again, it'll be down to people who are working on such things to guide you here.
- It's not clear to me what these constants are meant to be doing. The comment "Rotations matrix for 45 degree rotations" doesn't make sense to me in this context, because this is far too many constants to be such a matrix.
- If it is a rotation matrix, it is probably named [against our conventions](https://nubook.nubots.net/system/foundations/mathematics#rotations). These are confusing at first, but intuitive once you understand them. @ysims  can help you with naming matrices ;~)
- The biggest thing, is that these constants probably shouldn't even be in this file. They should ideally be read from a config/data file, like in the `.yaml` file [in this directory](https://github.com/NUbots/NUbots/tree/master/module/support/NUsight/data/config). Convert these into a yaml file, then read them from there.

controllers/teleport_controller/teleport_controller.cpp
```diff
@@ -0,0 +1,138 @@
...
+                                                  {0.286947, 0.661632, 0.692752, 2.60662},
+                                                  {1.69526e-09, 0.707103, 0.707111, 3.14159},
+                                                  {-0.286948, 0.661631, 0.692752, -2.60661},
+                                                  {-0.577352, 0.577347, 0.577352, -2.09441},
+                                                  {0.867739, -0.343277, -0.359429, 1.75753}};                               
+
+
+/* This is the main program of the supervisor. It creates an instance of the Supervisor
+ *     instance, launches its function(s) and destroys it at the end of the execution.
+ */
```
KipHamiltons, line 43: This is more the style which we would expect if you did a javadoc comment for this `main`
```suggestion
/** @brief Create a webots Supervisor, launch its functions, then destroy it
 */
```
Although "launch its functions" is quite vague, and doesn't really tell us much

controllers/teleport_controller/teleport_controller.cpp
```diff
@@ -0,0 +1,138 @@
...
+int main() {
+    std::cout << "a" << std::endl;
```
KipHamiltons, line 45: Why are we printing random letters to console?

controllers/teleport_controller/teleport_controller.cpp
```diff
@@ -0,0 +1,138 @@
...
+    // create the Supervisor instance and assign it to a robot
+    Supervisor* supervisor = new Supervisor();
```
KipHamiltons, line 47: Never use raw pointers (`new` and `delete`). Use smart pointers instead, or more conveniently, just do everything on the stack. Our codebase doesn't do much stuff with objects large enough to bother with heap allocations, so most of the time we just do everything on the stack, with the added bonus of performance.

controllers/teleport_controller/teleport_controller.cpp
```diff
@@ -0,0 +1,138 @@
...
+    Node* target           = supervisor->getFromDef("RED_1");
+    std::cout << "b" << std::endl;
```
KipHamiltons, line 49: Why are we printing random letters to console?

controllers/teleport_controller/teleport_controller.cpp
```diff
@@ -0,0 +1,138 @@
...
+    // Get the time step of the current world.
+    int timeStep = (int) supervisor->getBasicTimeStep();
+
+    std::cout << "c" << std::endl;
```
KipHamiltons, line 53: Why are we printing random letters to console?

controllers/teleport_controller/teleport_controller.cpp
```diff
@@ -0,0 +1,138 @@
...
+ *     instance, launches its function(s) and destroys it at the end of the execution.
+ */
+int main() {
+    std::cout << "a" << std::endl;
+    // create the Supervisor instance and assign it to a robot
+    Supervisor* supervisor = new Supervisor();
+    Node* target           = supervisor->getFromDef("RED_1");
+    std::cout << "b" << std::endl;
+    // Get the time step of the current world.
+    int timeStep = (int) supervisor->getBasicTimeStep();
```
KipHamiltons, line 51: Never use c-style casts. Use functional casts, static casts, or even reinterpret casts (reinterpret casts are usually ok for our purposes, despite what stackoverflow says, because we aren't doing security critical things or anything redhot which makes reinterpret casts a hot topic. @TrentHouliston has many more things to say about such things lol).

controllers/teleport_controller/teleport_controller.cpp
```diff
@@ -0,0 +1,138 @@
...
+
+    std::cout << "c" << std::endl;
+
+    // Add the motion manager to make the robot stand up and look around
+    managers::RobotisOp2MotionManager* mMotionManager = new managers::RobotisOp2MotionManager(supervisor);
```
KipHamiltons, line 56: Don't use raw pointers.

controllers/teleport_controller/teleport_controller.cpp
```diff
@@ -0,0 +1,138 @@
...
+    mMotionManager->playPage(54);
+
+    // Generate random seed
+    srand(time(NULL));
```
KipHamiltons, line 60: If we are generating a random seed, is it necessary to set one at all? (I am not sure)

controllers/teleport_controller/teleport_controller.cpp
```diff
@@ -0,0 +1,138 @@
...
+
+    // Main loop:
+    while (supervisor->step(timeStep) != -1) {
```
KipHamiltons, line 63: This loop condition confuses me. It should probably be reworked, or at least commented explaining why this exotic loop condition is necessary

controllers/teleport_controller/teleport_controller.cpp
```diff
@@ -0,0 +1,138 @@
...
+    std::cout << "c" << std::endl;
+
+    // Add the motion manager to make the robot stand up and look around
+    managers::RobotisOp2MotionManager* mMotionManager = new managers::RobotisOp2MotionManager(supervisor);
+    mMotionManager->playPage(54);
+
+    // Generate random seed
+    srand(time(NULL));
+
+    // Main loop:
```
KipHamiltons, line 62: This comment doesn't tell me anything really. What do you mean that it's the main loop? is this comment necessary?

controllers/teleport_controller/teleport_controller.cpp
```diff
@@ -0,0 +1,138 @@
...
+    while (supervisor->step(timeStep) != -1) {
+
+        // Grab the current translation field of the robot to modify
+        Field* target_translation_field = target->getField("translation");
+        // Convert the field to a vector to output to console
+        const double* target_translation_vec = target_translation_field->getSFVec3f();
+
+        // Output current location
+        std::cout << "Location: " << std::endl;
```
KipHamiltons, line 71: Logging these to a file is _probably_ preferred to printing them to console. Either that, or adding this information to the filename of the image we create with this field or something, or some other way of tying this information to the relevant image(s).

controllers/teleport_controller/teleport_controller.cpp
```diff
@@ -0,0 +1,138 @@
...
+         *     X ranges from -5.4 - 5.4, the positive value on the left hand side when looking
+         *     from red
+         *     Y ranges from -8 - 8, the positive value on the red goal side
+         *     Z should be set to 3.2 to be on the ground
+         *     If the robot teleports into an existing object it may run into issues for that
+         *     image only, after resetPhysics is should return to a regular state
+         */
+
+        double newPos[3];
+        newPos[0] = 5.4 - (rand() % 1080) / 100;  // target_translation_vec[0] + 1;
```
KipHamiltons, line 88: I don't understand what these end of line comments are saying

controllers/teleport_controller/teleport_controller.cpp
```diff
@@ -0,0 +1,138 @@
...
+        // Set new location
+        target_translation_field->setSFVec3f(newPos);
+
+        // Grab the current rotation field of the robot to modify
+        Field* target_rotation_field = target->getField("rotation");
+        // Convert the field to a vector to output to console
+        const double* target_rotation_vec = target_rotation_field->getSFRotation();
+
+        // Loop 4 times, one for each direction to face
+        for (int i = 0; i < rotationsSize; i++) {
```
KipHamiltons, line 101: Using a `std::array` or `std::vector` allows us to use range-based for loops. These are basically always preferred to manually setting dummy variables for for loops, like here.

controllers/teleport_controller/teleport_controller.cpp
```diff
@@ -0,0 +1,138 @@
...
+            // Output current rotation
+            std::cout << "Rotation: " << std::endl;
+            std::cout << "X: " << target_rotation_vec[0];
+            std::cout << " Y: " << target_rotation_vec[1];
+            std::cout << " Z: " << target_rotation_vec[2];
+            std::cout << " α: " << target_rotation_vec[3] << std::endl;
+
+            /* Prepare new rotation. These are saved in rotations matrix as the axis-angle
```
KipHamiltons, line 109: _Usually_ we use // comments instead of /* comments, with the exception of javadoc style /** comments.

controllers/teleport_controller/teleport_controller.cpp
```diff
@@ -0,0 +1,138 @@
...
+             *     manually move the robot motors to where we want them
+             */
+            mMotionManager->playPage(54);
+            // Other presets, comment out as desired
+            mMotionManager->playPage(15);
+            mMotionManager->playPage(1);
+        }
+    };
+
+    delete supervisor;
```
KipHamiltons, line 136: raw pointers bad. RAII is based

controllers/teleport_controller/teleport_controller.cpp
```diff
@@ -0,0 +1,138 @@
...
+            // This line is very important. The robot blows itself up after a 1:05 minutes otherwise
+            // Teleporting is strenuous work apparently
+            target->resetPhysics();
+
+            /* Make robot stand up and look around. This preset is slow and I'm not super
+             *     happy with it but it does the job for now. Below is a work in progress to
+             *     manually move the robot motors to where we want them
+             */
+            mMotionManager->playPage(54);
+            // Other presets, comment out as desired
```
KipHamiltons, line 130: should these be commented out? this comment makes it seem like they should before this is merged??

controllers/teleport_controller/teleport_controller.cpp
```diff
@@ -0,0 +1,138 @@
...
+            newRot[0] = rotations[i][0];
+            newRot[1] = rotations[i][1];
+            newRot[2] = rotations[i][2];
+            newRot[3] = rotations[i][3];
+
+            // Set new rotations
+            target_rotation_field->setSFRotation(newRot);
+
+            // This line is very important. The robot blows itself up after a 1:05 minutes otherwise
+            // Teleporting is strenuous work apparently
```
KipHamiltons, line 122: the robot blows itself up? what do you mean? this could be worth looking into - we don't want the robot "blowing up" 1.05 into a game lol :P

KipHamiltons (changes requested):
First things first - this is a really neat set of code. It's well documented, readable, and functional. Those things are very important, and a really good place to start.

Most of this review is superficial, and I haven't reviewed your approach to the actual problem (and I am not sure I am in a place to yet, having no experience with webots), but this review should be a good place to start, in terms of common practices for our code.
If this does the job, it's probably worth ignoring a lot of my "style" issues for now, and addressing things later, so that we can get stuff training asap, although that's not my call to make (visual mesh people would be best to decide that.)

controllers/teleport_controller/teleport_controller.cpp
```diff
@@ -0,0 +1,138 @@
...
+
+#include <RobotisOp2GaitManager.hpp>
+#include <RobotisOp2MotionManager.hpp>
+#include <webots/Supervisor.hpp>
+#include <iostream>
+#include <math.h>
+#include <time.h> 
+
+// All the webots classes are defined in the "webots" namespace
+using namespace webots;
```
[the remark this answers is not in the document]
Bidski: `using namespace std` is a horrendous thing and should be avoided at all costs. This, I believe, is mainly because many other libraries may implement functions in global namespace (horrible practice) that will clash with standard library functions.

I usually extend this to other single name `using`s (e.g. `using namespace eigen`). Mainly because it makes it more annoying to know where functions are coming from without really saving you much of anything.

If the namespace you are `using` is long and/or nested then using a `using` will make your code more readable, but we rarely, if ever, do this in a header file.

@TrentHouliston may have more thoughts on this

controllers/teleport_controller/teleport_controller.cpp
ysims: Seems to be a normal thing with supervisors and changing a robot's position/rotation
See https://www.cyberbotics.com/doc/reference/supervisor#wb_supervisor_node_reset_physics
If Cyberbotics put in supervisor stuff into the controller, which they very much might do for disentangling robots or similar, they will probably reset the robot's physics too - or we'll find that to be an issue in the mock matches, I assume.

controllers/teleport_controller/teleport_controller.cpp
ysims: https://cyberbotics.com/doc/guide/controller-programming
See part about `wb_robot_step`. 
The main loop is a specific thing in controllers in which each time step the controller code executes everything needed in that time step. Otherwise, without a loop, it'll just execute and stop, and the robot won't do anything. 
`step(timeStep)` is probably similar to that function in the docs, where the function takes `timeStep` seconds to execute and so gives a delay between each execution of the while loop. Why `-1` is another question, I guess that's what it returns when all is ok? Seems to be standard in webots.

controllers/teleport_controller/Makefile
```diff
@@ -0,0 +1,29 @@
...
+null :=
+space := $(null) $(null)
+WEBOTS_HOME_PATH=$(subst $(space),\ ,$(strip $(subst \,/,$(WEBOTS_HOME))))
+RESOURCES_PATH = $(WEBOTS_HOME)/projects/robots/robotis/darwin-op
+INCLUDE = -I"$(RESOURCES_PATH)/libraries/managers/include" -I"$(RESOURCES_PATH)/libraries/robotis-op2/robotis/Framework/include"
+LIBRARIES = -L"$(RESOURCES_PATH)/libraries/robotis-op2" -lrobotis-op2 -L"$(RESOURCES_PATH)/libraries/managers" -lmanagers
+CXX_SOURCES = $(wildcard *.cpp)
+
+### Do not modify: this includes Webots global Makefile.include
+include $(WEBOTS_HOME_PATH)/resources/Makefile.include
```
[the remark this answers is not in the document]
Lachourt: Gotcha, I'll be honest I don't really know how either makefiles or CMakeLists work so this is just a copypasta from another controller. I'll see if a copy from the NUgus controller will do the same thing for now, but defs something I need to learn how they work

controllers/teleport_controller/teleport_controller.cpp
```diff
@@ -0,0 +1,138 @@
+// File:          teleport_controller.cpp
+// Date:          23/03/2021
+// Description:   Method to teleport robot around the playing field for the purposes of
+//                    gathering training data
+// Author:        Lachlan Court
+// Modifications:
```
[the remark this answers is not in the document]
Lachourt: Is this license something I can just copypasta? Or is it specific to each file

controllers/teleport_controller/teleport_controller.cpp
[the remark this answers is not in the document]
Lachourt: Are you proposing I don't use the namespace in this way? This was just from the sample code in the webots repo and it worked so I kept it aha, I can work out which specific functions I'm using though and call webots::Motor or w/e if that would be more suitable

controllers/teleport_controller/teleport_controller.cpp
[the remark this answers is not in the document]
KipHamiltons: You should be able to copy it straight. just change the year to this year

controllers/teleport_controller/teleport_controller.cpp
Lachourt: What do you mean by we won't keep it? But yes I did know that and forgot to do it 🤦

controllers/teleport_controller/teleport_controller.cpp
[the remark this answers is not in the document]
KipHamiltons: Personally, I think that would be good

controllers/teleport_controller/teleport_controller.cpp
KipHamiltons: we probably won't keep it because its purpose is for a loop bound, and you should use range-based for loops for the standard containers (std::array, std::vector, etc.), or their `.size()` methods if you need the size for some reason. [.size() for std::vector on cppreference](https://en.cppreference.com/w/cpp/container/vector/size) (an incredible website btw)

controllers/teleport_controller/teleport_controller.cpp
Lachourt: The rotations in webots are in a format I've never seen before, 4 values, x y z range from -1 to 1 to indicate an axis of rotation that the 4th value, the angle should be applied to. Perhaps rotation matrix was the wrong term to use but it made sense to me, essentially I manually moved the robot around in the GUI to determine the rotations I wanted to use and copied them in here because it was too hard to calculate them on the fly (I'm sure it is possible but not being familiar with this method of handling rotation I was not confident delving into the maths of it all)

I am not sure how to use yaml files so I might need a hand with that but if that's the route i should be taking more than happy to look into it

controllers/teleport_controller/teleport_controller.cpp
Lachourt: Link is dead :( But yeah I am assuming that it would be similar to a python list having a size stored that I can access rather than having to save it separately

controllers/teleport_controller/teleport_controller.cpp
Lachourt: Ahaha again just a copypasta from the sample code, it worked so I kept it but can change it

controllers/teleport_controller/teleport_controller.cpp
Lachourt: Not knowing the exact way that the program was going to be used I have a couple of options in there that someone can comment out as they wish. They are meant to be uncommented but I wanted it to be clear that removing them wouldn't break the program if someone wanted to change the preset motions that the robot was performing

controllers/teleport_controller/teleport_controller.cpp
Lachourt: It was actually quite fun to watch, it would sit there perfectly still for about a minute and then just start tearing itself apart for no reason

controllers/teleport_controller/teleport_controller.cpp
KipHamiltons: sounds like they might using [unit quaternions](https://en.wikipedia.org/wiki/Quaternions_and_spatial_rotation) for rotation, but without naming it that? (I don't know enough about quaternions for rotations...) @ysims @Bidski is this a unit quaternion set up? [relevant docs page](https://cyberbotics.com/doc/reference/transform)

KipHamiltons: Calling them matrices is definitely not appropriate though IMO

controllers/teleport_controller/teleport_controller.cpp
KipHamiltons: Here's an example of a range based for loop in our codebase:
https://github.com/NUbots/NUbots/blob/master/module/behaviour/Controller/src/Controller.cpp#L62

It also uses the `auto` keyword, which is very handy to know about as well.

KipHamiltons: but yea, you've basically nailed it. It's a lot like `len(list)` in python, although it's a little more like if each list had a `.len()` method instead.

controllers/teleport_controller/teleport_controller.cpp
KipHamiltons: It's not that the supervisor object shouldn't be deleted, it's that smart pointers do the deletion for us when they go out of scope. Calling delete explicitly means you've used raw pointers (basically any time you use the `new` keyword, it's a raw pointer), and raw pointers are just a fast track to memory leaks and other fun issues.

controllers/teleport_controller/teleport_controller.cpp
KipHamiltons: That makes some sense to me. I think that some of that info could be added to the comment though

KipHamiltons: or I suppose, ideally this would be a config value in a config file, so you know that this is something which changes how the thing runs, and where to change it. The description you just gave would be condensed to stuff relevant to someone changing the config file later.

controllers/teleport_controller/teleport_controller.cpp
KipHamiltons: Ahh that's what that was 😂

controllers/teleport_controller/teleport_controller.cpp
Bidski: Sounds like [AxisAngle](https://en.m.wikipedia.org/wiki/Axis%E2%80%93angle_representation) tbh.

controllers/teleport_controller/teleport_controller.cpp
Bidski: Also, use the c++ random number library.

controllers/teleport_controller/teleport_controller.cpp
KipHamiltons: Magic numbers in general are bad ([see this wiki page about "unnamed numerical constants" for more info on what a magic number is](https://en.wikipedia.org/wiki/Magic_number_(programming))). Magic numbers in config files with descriptions/comments/explanations (which I guess means that they wouldn't be magic numbers any more by definition lol) are much better.

controllers/teleport_controller/teleport_controller.cpp
Bidski: Have a look in the networking PR #1 for how I used a `std::unique_ptr` (a type of smart pointer) in that controller.

controllers/teleport_controller/teleport_controller.cpp
KipHamiltons: Ahh true. I thought that was an eigen-only thing lol. that looks like exactly what we're looking for

Lachourt: Yeah I think axis angle is the one, it also has an option to view the rotation in quarternions but I don't know how to set it as quarternions

controllers/teleport_controller/teleport_controller.cpp
Lachourt: Yep okidoki

controllers/teleport_controller/teleport_controller.cpp
Lachourt: Yeah alright I'll have a look

controllers/teleport_controller/teleport_controller.cpp
Lachourt: Are you proposing having it read from a config file which you can edit? Or just have a txt that explains what's going on?
And yes silly oversight on the numbers thing, I had the reference open in front of me so I knew exactly what was going on 😆

controllers/teleport_controller/teleport_controller.cpp
Lachourt: Yep okay

controllers/teleport_controller/teleport_controller.cpp
Lachourt: Ummmmm just in case we forget the alphabet? std::cout > debugger :laughing:

controllers/teleport_controller/teleport_controller.cpp
Lachourt: I just like how b looks as the program boots up

controllers/teleport_controller/teleport_controller.cpp
Lachourt: My last name starts with it, I like it

controllers/teleport_controller/teleport_controller.cpp
Lachourt: Not sure what this means sorry, that's the only way I've ever done a typecast

controllers/teleport_controller/teleport_controller.cpp
Lachourt: Depends, if we only ever use the default seed it would run the same every time we started it up wouldn't it? In the interests of getting a broad data set I thought that would be best
I didn't realise there was a random number library, I'll look into it

controllers/teleport_controller/teleport_controller.cpp
TrentHouliston: Basically if it is a constant and has meaning, stick it in a variable and give the variable a name

controllers/teleport_controller/teleport_controller.cpp
Lachourt: Yep this was just the format of the sample code, it runs everything in that loop repeatedly until the simulation is terminated

controllers/teleport_controller/teleport_controller.cpp
Lachourt: Leftover from sample, will remove

controllers/teleport_controller/teleport_controller.cpp
Lachourt: Happy to change this, was just doing console so I could monitor is as I tested it but wasn't sure what the end game was so just left it as it was

controllers/teleport_controller/teleport_controller.cpp
Lachourt: Leftover comment from testing, will remove

controllers/teleport_controller/teleport_controller.cpp
TrentHouliston: Just a quick mention about one of the easiest reasons to see when a raw pointer will do a memory leak

```
int* pointer = new int;
if(i_feel_like_it) {
    throw std::runtime_error();
}
delete pointer;
```

In this case you have a memory leak because you never delete the pointer (you throw out of the function first)
If it were a smart pointer, it would be deleted

controllers/teleport_controller/Makefile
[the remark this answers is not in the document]
Lachourt: Okay I've created a CMakeLists.txt and changed the names but webots still won't compile it without a makefile so I'm not sure how to test it lol

controllers/teleport_controller/teleport_controller.cpp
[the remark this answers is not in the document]
TrentHouliston: Using namespace is almost always bad, there are some very very few use cases where it's a good idea. But if you are good enough to know when those are, you probably don't need c++ tips so much anymore :P

Basically, the problem with it is that it drags symbols into your current namespace that can conflict with symbols you're defining yourself. E.g. if you wrote yourself a nice function named `count`, and had `using namespace std` the compiler wouldn't know if you wanted to refer to your version of `count`, or the standard libraries version of `count`. The error it gives you, in this case, is very hard to read if you haven't read it before and can mess up a few hours of your time.

This is especially bad when you do using namespace in headers because it doesn't just pollute your code you're writing now, but potentially anyone who includes that header, even transitively. So you can add using namespace in your header happily coding away and a weird esoteric error can crop up on the other side of the codebase with a message like "Ambiguous symbol here!` and when you are causing compile errors in code you're not even writing you'll have a bad day (or potentially whoever includes your header in the future will.

Remember kids, always code as if the person maintaining your code is a violent psychopath who knows where you live. Because they will be, and it'll probably just be you a few months from now.

Hierarchy of bad using/using namespace ideas

1. 😭 `using namespace` in a header
  - never do this you monster, you're screwing over yourself and everyone else!
2. 😢 `using namespace` in a cpp file 
  - better than doing this in a header so at least you're not screwing over others  but you're still screwing over yourself and future people who work on this code
3. 😞 `using x = ...` in a header
  - don't do this unless you really want others to access the thing you're using in other files like this
5. 😐 `using namespace` in a function 
  - This is the one time that I won't yell at you for doing a using namespace it doesn't pollute the scope so much (only that function is affected) and nobody likes typing `std::chrono::duration_cast<std::chrono::duration<double>(std::chrono::steady_clock::now()).count()` when they can do `duration_cast<duration<double>>(steady_clock::now()).count();` just with a little `using namespace std::chrono;`
4. 😃 `using x = ...` in a cpp file 
  - this one is super useful and we do it all the time, go nuts

controllers/teleport_controller/teleport_controller.cpp
Lachourt: Yeah okay that makes heaps of sense, cool, I'll probably just change it around so the whole thing runs on the stack lol, it's fairly lightweight and that way I don't need to worry about it at all

controllers/teleport_controller/teleport_controller.cpp
Lachourt: Would this be something I should use a #define for or just a regular constant is okay?

TrentHouliston: You should ever really be using #define in c++, there are very few situations where it's a good idea. The main reason being they don't have type information and are just doing text replacement so you can get into some weird spots.

In c++ we use `constexpr` to define constants with a type like

```cpp
constexpr int ULTIMATE_ANSWER = 42;
```

controllers/teleport_controller/teleport_controller.cpp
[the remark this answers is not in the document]
Lachourt: Awesome thanks for that! I'll keep it in mind :)

controllers/teleport_controller/teleport_controller.cpp
Lachourt: Yeah okay sweet, ty

controllers/teleport_controller/Makefile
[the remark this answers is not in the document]
KipHamiltons: the process is a bit different. You'll have to compile it yourself, then put the binary in a specific spot IIRC. @ysims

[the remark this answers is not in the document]
ysims: If you take a look at the update-readme branch there are some instructions (Linux). The binary will be put in the right spot, if everything is working properly.

controllers/teleport_controller/teleport_controller.cpp
KipHamiltons: [Relevant cppreference page](https://en.cppreference.com/w/cpp/language/explicit_cast). Typecasts can be a bit of a rabbit hole, but it's interesting enough. 

Here are some heuristics for casting in our codebase:

- `(type) thing` is a c-style cast. These are bad
- `type(thing)` is a functional cast. these are fine
- `static_cast<type>(thing)` is a static cast. these are fine
- `reinterpret_cast<type>(thing)` is a reinterpret cast. People online will yell at you for using these, but not for reasons we care about in our code

controllers/teleport_controller/teleport_controller.cpp
KipHamiltons: well it depends what the default behaviour is, as in, what happens if we don't give it a seed at all? most programs that can be seeded use a random seed if you don't give them a seed explicitly. It's fine to seed it anyway of course.

controllers/teleport_controller/teleport_controller.cpp
TrentHouliston: Yeah, a c style cast is a sledgehammer it will do a static cast and a reinterpret cast at the same time. This is bad because we really don't want to be using reinterpret casts by accident.

Also, it's harder with a c style cast to know what the cast is applying to
`int(a) + b` is much easier to understand that the a is getting cast to an int rather than `(int) a + b` where it's the same, but harder to tell that it's a that's cast, and not `a + b`

Lachourt: Wait int(a) works in c++?? Ahhhhhhh
I do that in python but I didn't realise you could do in c++ as well 😭

controllers/teleport_controller/teleport_controller.cpp
Lachourt: I will run a few tests and see if it makes a difference, if not I'll remove it

controllers/teleport_controller/teleport_controller.cpp
TrentHouliston: I mean technically what you're doing is constructing a new `int` instance using the value you pass in as constructor arguments. But it basically works as a cast

controllers/teleport_controller/teleport_controller.cpp
TrentHouliston: Axis angle and unit quaternions are quite similar given the fact that they are all from the range -1 to 1. Axis angle the angle part would often be greater than or less than -1 to 1. An easy way to check is to see if the 2 norm of the values is 1.
If we are using Eigen in here and this is one of our constants, you should be storing these in one of Eigen's types (AngleAxisd or Quaterniond at least if not Affine3d`

controllers/teleport_controller/teleport_controller.cpp
```diff
@@ -0,0 +1,138 @@
...
+               calculation is rough to calculate on the fly*/
+
+            double newRot[4];
+            newRot[0] = rotations[i][0];
+            newRot[1] = rotations[i][1];
+            newRot[2] = rotations[i][2];
+            newRot[3] = rotations[i][3];
+
+            // Set new rotations
+            target_rotation_field->setSFRotation(newRot);
```
Bidski, line 119: `rotations[i]` is already an array of 4 doubles, so you dont need to make a new array and copy the values across.
```suggestion
            // Set new rotations
            target_rotation_field->setSFRotation(rotations[i]);
```

controllers/teleport_controller/teleport_controller.cpp
Bidski: > For every different seed value used in a call to srand, the pseudo-random number generator can be expected to generate a different succession of results in the subsequent calls to rand.
> 
> Two different initializations with the same seed will generate the same succession of results in subsequent calls to rand.

In other words, setting the seed to a specific value will effectively guarantee the same random sequence of numbers. Setting the seed to a different value (like the current time) will ensure you get a different sequence of random numbers every time you run the program (i.e. every time you run you should expect different sequences of random behaviour)

controllers/teleport_controller/teleport_controller.cpp
Lachourt: I've never heard of those before aha I'll confirm which one it is and swap it over. Given that there are a number of different rotations, should I be storing each of those Eigen types in an std::vector? Am I understanding that correctly?

Bidski (changes requested):

controllers/teleport_controller/teleport_controller.cpp
Bidski: Yea, make an `std::vector<Eigen::AngleAxisd>` or a `std::array<Eigen::AngleAxisd, N>` depending on the context (where `N` is the number of elements in the array, known at compile-time)

TrentHouliston: If you know how many there are and it won't change, use `std::array`, if you will be adding/removing elements to it then use `std::vector`

controllers/teleport_controller/teleport_controller.cpp
KipHamiltons: doing some reading... why the f does C++ rand automatically use `seed == 1` without an `srand()` call wtf ahaha. Yea, you'll need to seed it for sure Lachlan

TrentHouliston: Well, if you're using c++11 or higher, you shouldn't ever be using std::rand anyway. It has problems.

Modern method is to use the new c++ random functionality to do this.

Pick a distribution and then do what it does below, make a pseudorandom number generator based off a true random source.
https://en.cppreference.com/w/cpp/numeric/random/uniform_int_distribution

controllers/teleport_controller/teleport_controller.cpp
Lachourt: Gotcha, probs do an array and I can have a size in the yaml config file

TrentHouliston: If your size is coming from a yaml file it's dynamic because you don't know it without running the code

controllers/teleport_controller/teleport_controller.cpp
Lachourt: Ohhhhh okay okay neat that makes sense

controllers/teleport_controller/teleport_controller.cpp
Lachourt: My only experience with c++ is 98 thanks to SENG1120 😆  I haven't delved into the wonders that are c++<literally anything else> so I'm excited to learn this better way of doing random numbers!

controllers/teleport_controller/teleport_controller.cpp
Lachourt: True true, I'll be changing it into an array of eigen rotations instead of a 2D array so this will be sorted once I've done that anyway

controllers/teleport_controller/teleport_controller.cpp
Lachourt: I think I have misunderstood how a yaml file works then, I thought I would be able to read the size, create the array according to that size, and then read the rotations and add them into the array?

Bidski: To use an `std::array` you need to know exactly how many elements will be stored in it at compile-time. If the size of the array is stored in a YAML file then you wont know the size of the array until run-time, so you will have to use a `std::vector` instead.

KipHamiltons: there's nothing special about `yaml` files. That's just the format we use for config files - it's basically formatted plaintext. The way we (usually) read the data from them is by reading them at run time, with NUClear emitting a `Configuration` message. The module reading the `yaml` gets a config object it can use, in an `on<Configuration>(..)` call. It uses that config object to assign the config values to member variables of the module. [Random example from the codebase](https://github.com/NUbots/NUbots/blob/master/module/motion/QuinticWalk/src/QuinticWalk.cpp#L68).

In this case, since it's not a module (or more accurately, a NUClear reactor), we don't have that formal configuration process with the `on` statement. Instead, you'll probably want to use the library which does it for us directly: [yaml-cpp](https://github.com/jbeder/yaml-cpp). 

A big benefit to doing this generally is that we can tweak the parameters to change how the program behaves without going deep into the trenches to find what we should change. We won't end up wondering if we're changing the right thing, or if that change will fundamentally break the program (hopefully). A similar effect can be realised by naming magic numbers generally (although this case isn't quite a magic number, but more a set of magic tuples), but consistently using config/data files makes it easy.

KipHamiltons:
Now that #10 is merged, you could consider using that script to generate the controller. That'll avoid the headaches of writing a CMakeLists.txt for the first time and let you focus on the logic of your controller.

I don't mean to say you should - just letting you know it's an option.

ysims (commented):
You'll want to add back in the `CMakeLists.txt` file otherwise it doesn't compile. Also add `/controllers/teleport_controller/teleport_controller` to the `.gitignore` file so no one accidentally commits the binary.

Lachourt:
> You'll want to add back in the `CMakeLists.txt` file otherwise it doesn't compile. Also add `/controllers/teleport_controller/teleport_controller` to the `.gitignore` file so no one accidentally commits the binary.

Where's the gitignore file?

Also whoops about the CMakeLists but I don't actually know if it worked anyway, that one was generated from the auto generator script and it didn't seem to include everything. I've been using the builtin makefile from webots

controllers/teleport_controller/teleport_controller.cpp
```diff
@@ -0,0 +1,147 @@
...
+            // Output current rotation
+            try {
+                std::ofstream log;
+                log.open("log.txt", std::fstream::app);
+                log << "Rotation: " << std::endl;
+                log << "X: " << target_rotation_vec[0];
+                log << " Y: " << target_rotation_vec[1];
+                log << " Z: " << target_rotation_vec[2];
+                log << " α: " << target_rotation_vec[3] << std::endl;
+                log.close();
```
KipHamiltons, line 118: I am wondering if reopening this file will end up being a bottleneck for running this? probably not, since this generates, renders and saves a training image each iteration... but could be worth seeing if keeping the file open gives you a noticeable bump in performance (if we are finding this script is too slow later).

controllers/teleport_controller/teleport_controller.cpp
```diff
@@ -0,0 +1,147 @@
...
+            }
+            catch (const std::string& e) {
```
KipHamiltons, line 120:
```suggestion
            catch (const YAML::BadFile& e) {
```

controllers/teleport_controller/teleport_controller.cpp
```diff
@@ -0,0 +1,147 @@
...
+                std::cout << e << std::endl;
```
KipHamiltons, line 121:
```suggestion
                std::cerr << e << std::endl;
```

controllers/teleport_controller/teleport_controller.cpp
```diff
@@ -0,0 +1,147 @@
...
+            }
+
+            // Prepare new rotation. These are saved in rotations vector as the axis-angle
+            // calculation is rough to calculate on the fly
+
+            // The setSFRotation webots function expects an array of doubles, not a vector
+            double newRot[4];
```
KipHamiltons, line 128: std::array is based
```suggestion
            std::array<double, 4> newRot = rotation;
```

controllers/teleport_controller/teleport_controller.cpp
```diff
@@ -0,0 +1,147 @@
...
+            for (long unsigned int j = 0; j < rotations[i].size(); j++) {
+                newRot[j] = rotations[i][j];
+            }
```
KipHamiltons, line 131:
```suggestion
```

controllers/teleport_controller/teleport_controller.cpp
```diff
@@ -0,0 +1,147 @@
...
+
+            // Apply new rotation and reset physics to avoid robot tearing itself apart
+            target_rotation_field.setSFRotation(newRot);
```
KipHamiltons, line 134: To get the underlying C array from the std::array
```suggestion
            target_rotation_field.setSFRotation(newRot.data);
```

controllers/teleport_controller/teleport_controller.cpp
```diff
@@ -0,0 +1,147 @@
...
+            target.resetPhysics();
+
+            // The following performs a number of preset movements/routines loaded from the
+            // config file that the robot can carry out. Any of them can be removed or more
+            // added but leave at least one otherwise there will be no delay between each
+            // step of this loop
+            for (int i : playPages) {
```
KipHamiltons, line 141:
```suggestion
            for (const int& i : playPages) {
```

controllers/teleport_controller/teleport_controller.cpp
```diff
@@ -0,0 +1,147 @@
...
+        // Y ranges from -8 - 8, the positive value on the red goal side
+        // Z should be set to 0.32 to be on the ground
+        // If the robot teleports into an existing object it may run into issues for that
+        // image only, after resetPhysics is should return to a regular state
+
+
+        double newPos[3];
+        newPos[0] = 5.4 - xDistrib(gen) / 100;
+        newPos[1] = 8 - yDistrib(gen) / 100;
+        newPos[2] = 0.32;
```
KipHamiltons, line 97:
```suggestion
        std::array<double, 3> newPos;
        newPos[0] = 5.4 - xDistrib(gen) / 100.0;
        newPos[1] = 8 - yDistrib(gen) / 100.0;
        newPos[2] = 0.32;
```

controllers/teleport_controller/teleport_controller.cpp
```diff
@@ -0,0 +1,147 @@
...
+            std::ofstream log;
+            log.open("log.txt", std::fstream::app);
+            log << "Location: " << std::endl;
+            log << "X: " << target_translation_vec[0];
+            log << " Y: " << target_translation_vec[1];
+            log << " Z: " << target_translation_vec[2] << std::endl;
+            log.close();
+        }
+        catch (const std::string& e) {
+            std::cout << e << std::endl;
```
KipHamiltons, line 81:
```suggestion
            std::cerr << e << std::endl;
```

controllers/teleport_controller/teleport_controller.cpp
```diff
@@ -0,0 +1,147 @@
...
+        try {
+            std::ofstream log;
+            log.open("log.txt", std::fstream::app);
+            log << "Location: " << std::endl;
+            log << "X: " << target_translation_vec[0];
+            log << " Y: " << target_translation_vec[1];
+            log << " Z: " << target_translation_vec[2] << std::endl;
+            log.close();
+        }
+        catch (const std::string& e) {
```
KipHamiltons, line 80:
```suggestion
        catch (const YAML::BadFile& e) {
```

controllers/teleport_controller/teleport_controller.cpp
```diff
@@ -0,0 +1,147 @@
...
+std::vector<std::vector<double>> rotations;
+// Preset motions for the robot will be read and saved here
+std::vector<int> playPages;
+
+int main() {
+
+    // Load config file
+    try {
+        YAML::Node config = YAML::LoadFile("config.yaml");
+        rotations         = config["rotations"].as<std::vector<std::vector<double>>>();
```
KipHamiltons, line 38:
```suggestion
        rotations         = config["rotations"].as<std::vector<std::array<double, 4>>>();
```

controllers/CMakeLists.txt
```diff
@@ -1,2 +1,2 @@
 add_subdirectory(nugus_controller)
-
+add_subdirectory(teleport_controller)
```
KipHamiltons, line 2: Add newline at end of file to appease the gods
```suggestion
add_subdirectory(teleport_controller)

```

.gitignore
```diff
@@ -3,5 +3,5 @@ worlds/.*.wbproj
 build/
 
 /controllers/nugus_controller/nugus_controller
-
-/nugus_controller_test
+/controllers/teleport_controller/teleport_controller
+/nugus_controller_test
```
KipHamiltons, line 7: Add newline at end of file to appease the gods
```suggestion
/nugus_controller_test

```

KipHamiltons (changes requested):
Mostly C++ stuff really. I can't comment too much on the actual approach and all of that, since I don't know anything about the webots controllers and stuff, so we'll probably want someone who can help with that stuff to look at it too. Testing on linux and windows before approval is to be done too of course

controllers/teleport_controller/teleport_controller.cpp
```diff
@@ -0,0 +1,147 @@
...
+        playPages         = config["playpages"].as<std::vector<int>>();
+    }
+    catch (const std::string& e) {
```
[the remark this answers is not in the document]
Lachourt: It does the same either way you get error message for both but I've changed it to specify BadFile and ParserException errors

controllers/teleport_controller/teleport_controller.cpp
```diff
@@ -0,0 +1,147 @@
...
+        std::cout << e << std::endl;
+        return 0;
```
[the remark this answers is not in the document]
Lachourt: True

controllers/teleport_controller/teleport_controller.cpp
```diff
@@ -0,0 +1,147 @@
...
+    // Generate random seed
+    std::random_device rd;   // Will be used to obtain a seed for the random number engine
+    std::mt19937 gen(rd());  // Standard mersenne_twister_engine seeded with rd()
+    std::uniform_int_distribution<> xDistrib(0, 1080);
+    std::uniform_int_distribution<> yDistrib(0, 1600);
+
+    while (supervisor.step(timeStep) != -1) {
+
+        // Grab the current translation field of the robot to modify
+        webots::Field& target_translation_field = *target.getField("translation");
```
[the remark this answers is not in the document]
Lachourt: The getField() webots function returns a pointer to a field object, which I'm then dereferencing and saving in the target_translation_field reference as a memory address to the object. As far as I understand :sweat_smile:

controllers/teleport_controller/teleport_controller.cpp
```diff
@@ -0,0 +1,147 @@
...
+        playPages         = config["playpages"].as<std::vector<int>>();
+    }
+    catch (const std::string& e) {
+        std::cout << e << std::endl;
+        return 0;
+    }
+
+    // create the Supervisor instance and assign it to a robot
+    webots::Supervisor supervisor = webots::Supervisor();
+    webots::Node& target          = *supervisor.getFromDef("RED_1");
```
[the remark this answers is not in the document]
Lachourt: See above

controllers/teleport_controller/teleport_controller.cpp
```diff
@@ -0,0 +1,147 @@
...
+#include <RobotisOp2MotionManager.hpp>
+#include <fstream>
+#include <iostream>
+#include <random>
+#include <webots/Supervisor.hpp>
+
+#include "yaml-cpp/yaml.h"
+
+// AxisAngle rotations will be read from a config file and saved here
+std::vector<std::vector<double>> rotations;
```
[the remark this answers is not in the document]
Lachourt: Gotcha

controllers/teleport_controller/teleport_controller.cpp
Lachourt: I'll check it out

controllers/teleport_controller/teleport_controller.cpp
Lachourt: I don't even need to declare a new one if I'm using a for each loop so just got rid of this altogether

Lachourt: Using the for each loop for the rotations I don't even need to have this variable or the for loop after it

controllers/teleport_controller/teleport_controller.cpp
Lachourt: Removed this section already

controllers/teleport_controller/teleport_controller.cpp
Lachourt: This isn't a YAML file tho, just a txt. Is it still the same exception?

Lachourt: This isn't a YAML file tho, just a txt. Is it still the same exception?

controllers/teleport_controller/teleport_controller.cpp
```diff
@@ -0,0 +1,135 @@
...
+    }
+    catch (const YAML::ParserException& e) {
+        std::cerr << e.msg << std::endl;
+        return 2;
+    }
+    
+
+    // create the Supervisor instance and assign it to a robot
+    webots::Supervisor supervisor = webots::Supervisor();
+    webots::Node& target          = *supervisor.getFromDef("nugus");
```
ysims, line 50: This shouldn't apply to just one robot in the scene, it should be able to be applied to any robot and work.

controllers/teleport_controller/teleport_controller.cpp
```diff
@@ -0,0 +1,135 @@
...
+          
+
+            // Prepare new rotation. These are saved in rotations vector as the axis-angle
+            // calculation is rough to calculate on the fly
+
+            // Apply new rotation and reset physics to avoid robot tearing itself apart
+            target_rotation_field.setSFRotation(rotation.data());
+            target.resetPhysics();
+            
+
```
KipHamiltons, line 130: whitespace is too much

controllers/teleport_controller/teleport_controller.cpp
```diff
@@ -0,0 +1,135 @@
...
+        for (const std::array<double, 4>& rotation : rotations) {
+            // Output current rotation
+            
+            std::ofstream log;
+            log.open("teleport_controller_log.txt", std::fstream::app);
+            if (!log.good())
+            {
+              std::cout << "Error writing to log file" << std::endl;
+            }
+            log << "Rotation: " << std::endl;
```
KipHamiltons, line 115: Add timestamp to log please

controllers/teleport_controller/teleport_controller.cpp
```diff
@@ -0,0 +1,135 @@
...
+        webots::Field& target_rotation_field = *(target.getField("rotation"));
+        // Convert the field to a vector to output to console
+        const double* target_rotation_vec = target_rotation_field.getSFRotation();
+        
+        // Loop once for each rotation
+        for (const std::array<double, 4>& rotation : rotations) {
+            // Output current rotation
+            
+            std::ofstream log;
+            log.open("teleport_controller_log.txt", std::fstream::app);
```
KipHamiltons:
```suggestion
            log.open("teleport_controller.log", std::fstream::app);
```

controllers/teleport_controller/teleport_controller.cpp
ysims: std::unique_ptr<Robot> nugus = std::make_unique<Robot>();

worlds/kid.wbt
```diff
@@ -32,101 +32,40 @@ DEF RED_1 Darwin-op {
   hidden position_0_4 -0.844739262486803
```
KipHamiltons, line 32: unchange this file please

KipHamiltons (changes requested):

ysims (approved):
LGTM

Lachourt:
Just to make an even 150 comments :D

PR #16, opened 2021-04-07
Fix variable initialisation bug

Bidski:


controllers/nugus_controller/nugus_controller.cpp
```diff
@@ -133,9 +133,25 @@ int main(int argc, char** argv) {
         return 0;
```
JosephusPaye, line 133: Does returning non-zero here make sense since there was an error? (same with other similar returns below).

Bidski: What error code would you suggest

JosephusPaye: `1`, or better yet, [`EXIT_FAILURE`](https://stackoverflow.com/a/13667390/5800506)

ysims (approved):
It works! Tested on Windows.

JosephusPaye (approved):
LGTM

PR #18, opened 2021-04-08
Update NUgus model kinematics

Tom0Brien:
This PR updates the kinematics of the model to better align with the code base.

TODO:

- [x] Update leg joints coordinates and positions
- [ ] Fix hip roll rotation (currently axis does not pass through the visual servo axis)
- [ ] Update rest of body, torso, shoulders, head and arms
- [x] Add inertial properties for each link which match the codebase


ysims:
@Tom0Brien this might be worth closing. Particularly with the changes happening to the real robot, we'll likely want to export them as urdf and go through the tool, which will undo any work here.

PR #19, opened 2021-04-09
Updates the field with the new Webots RoboCup field

ysims:
Cyberbotics updated the RoboCup field a lil bit ago. This updates our stuff with their stuff.


worlds/adult.wbt
```diff
@@ -1,74 +1,108 @@
...
+#VRML_SIM R2021b utf8
 WorldInfo {
   info [
     "Description: official soccer simulation for the 2021 Robocup Virtual Humanoid League (adult size)"
-    "Version 0.1"
+    "Version 0.2"
   ]
   title "Robocup V-HL Adult"
-  basicTimeStep 16
+  basicTimeStep 8
```
Lachourt, line 8: What's the benefit of changing the timestep? Or is that just the standard for robocup?

Lachourt (commented):
Is there a reason we're still keeping the adult world up to date if we're only entering the kid league? Or have I misunderstood something

worlds/adult.wbt
ysims: Physics will update faster. I think it is the standard, or at least the conditions we'll be playing under (I don't think we're meant to get anymore field changes). It's not that necessary but may be useful if we want to play under their conditions.

ysims:
> Is there a reason we're still keeping the adult world up to date if we're only entering the kid league? Or have I misunderstood something

I've been thinking about this. I'm tempted to keep it because we do want to compete in the future in adult size, but the field might be different by then anyway...

worlds/adult.wbt
Lachourt: Fair enough, a smaller timestep will be great for vision data collection anyway so

Lachourt:
LGTM but I can't approve 👍

Lachourt (approved):
Wait no I can approve 😀

PR #20, opened 2021-04-13
Patched Torso Hole

AlanaNoonan:


ysims (approved):
Can confirm torso is patched up. 
I've checked the diff in GitKraken and it looks like the other lines are just removing spaces at the end. Might be a Windows thing or a formatting tool thing, but it's not really a problem.

PR #22, opened 2021-04-14
Add motor parameters

TheCoolKidsTable:
This PR adds:

-  Parameters for the MX106 and MX64 servos and assigns them to the correct motors. (MX106 for hip,knee,ankles and MX64 for arms/neck)
- PID gains for the motors (Obtained from Simscape)
- Sets selfCollision field for robot to TRUE

TODO:

- [x]  Add backlash


protos/NUgus.proto
```diff
@@ -1243,4 +1314,4 @@ PROTO NUgus [
       ]
     }
   }
-}
+}
```
ysims, line 1317: newline? :smile: Otherwise looks good

ysims (commented):
Might need someone who can review the actual values :smile:

xXBenjaminYoungXx:
Add comments on which Dynamixel model is being represented on the Rotational Motor Object.

Bidski (changes requested):
I think you should move the `boundingObject` changes to a separate PR. 

As for the backlash, I think we might find that they are expecting this as a measure of radians (because it is a HingeJoint). In which case 0.33 degrees (from the docs) is 0.005759587 radians (0.006). Have we tested with this value? Ludovic does warn against using a value that is too small

Bidski (approved):
LGTM

Bidski (dismissed):
Seems ok to me

ysims (approved):
LGTM

PR #24, opened 2021-04-14
Change syntax to match NUbots requirements

Peter-Salem:
Robot Model Specifications, under "structure of the robot", need to annotate body parts with suffixes, hence "right shoulder pitch" becomes "right shoulder pitch [shoulder]"


Peter-Salem:
also only changed one to make sure its correct

TheCoolKidsTable (approved):
LGTM!

PR #26, opened 2021-04-23
Add touch sensors to NUgus feet

NDxiura:
8 Touch sensors to NUgus feet

### DIFF
--- a/protos/NUgus.proto
+++ b/protos/NUgus.proto
@@ -652,6 +652,114 @@ PROTO NUgus [
                                               creaseAngle 1
                                             }
                                           }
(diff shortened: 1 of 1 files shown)


ysims:
I'm having trouble running this. Something is changing the proto when I load the world, so one of the touch sensors is in the wrong place (but the diff here looks ok and has a different values). Could you redo this PR as a branch on NUbots instead of the fork? You should be able to push now that you've been added to the organisation on GitHub.

NDxiura:
I think I know what causes that, probably a thing in the world file but I
need to figure out why it happens. Will fix and upload to github tomorrow.

PR #27, opened 2021-04-23
Add simplified collision boxes

TheCoolKidsTable:
This PR adds simplified collision boxes for the robot model. This was achieved using the tool provided by bitbots https://github.com/bit-bots/simplify_urdf_collision.

**Optimal bounding boxes generated:**
image


xXBenjaminYoungXx:
Needs to be tested (Perhaps use of simple controller). I might also add more comments where necessary then good to go!

JosephusPaye (approved):
I think we can merge now. Comments can come in a follow up PR.

worlds/kid.wbt
```diff
@@ -89,6 +89,9 @@ DEF BLUE_1 NUgus {
   rotation 0 0 1 3.1415
   controller "nugus_controller"
   controllerArgs [
-    "32"
+    # TCP port to listen on
```
ysims, line 92: Why is this in this PR?

ysims (changes requested):
@Bidski put in extra things maybe in the wrong branch?

worlds/kid.wbt
JosephusPaye: Was needed to test, and it doesn't seem to be in main.

JosephusPaye: I think it would be easier to add here and merge. But we could move it to a separate PR.

ysims (approved):
I guess 🤷

PR #28, opened 2021-04-23
Fixes include hack I did yesterday

KipHamiltons:
yesterday I just included the full path of the eigen deps to get it working. This is what I should have done.

Making a PR so that you can make sure it works on your machine before merging to that branch @ysims


ysims (approved):
Ye this doesn't break Eigen when I try it.

PR #30, opened 2021-04-24
added touch sensors to the feet of NUgus.proto

NDxiura:


ysims (commented):
Ah some copy and paste fun 😅  - check out the 'files changed' tab and revert the removed stuff. Might be easiest to copy the two touch sensor blocks of code somewhere, copy back in the current `main` version of `NUgus.proto`, and then copy back in the two touch sensor blocks.

moosahassanx:
Bit hard to follow without any comments.

Issue #17, opened 2021-04-08
Look for client connections in a non-blocking manner in the controllers


Bidski:
Currently, the controller will block waiting for a client to connect to it. We should be able to use `select()` to check for new client connections without blocking if there are none. This will also allow us to have multiple clients connected to a single controller (not sure if this is an important feature or not).

[This forum post](https://www.linuxquestions.org/questions/programming-9/asynchronous-accept-with-posix-sockets-891070/#post4411955) gives an idea of how to achieve this.

TrentHouliston:
`poll` > `select`

ysims:
I think this is how the player controller (which we've taken for the nugus_controller) works. But we should make it tidier.

PR #31, opened 2021-04-26
Adds receiving ActuatorRequests and sending SensorMeasurements

ysims:
Mostly taken from cyberbotics/webots#2888. Removed RoboCup-specific stuff. Changed some of the code. 

It would be great if someone could look at the code because I'm sure it can be made better. And I've never properly used smart pointers before so I wouldn't be surprised if I've messed things up there. 

I don't think image compression is finished so I've commented it out.


controllers/nugus_controller/nugus_controller.cpp
```diff
@@ -55,76 +81,317 @@ class NUgus : public webots::Robot {
...
+        }
+    }
+
+    void sendData() {
+        // Create the SensorMeasurements message
+        auto sensorMeasurements = std::make_unique<SensorMeasurements>();
 
-                // Send a message to the client
-                msg.set_num(current_num);
+        //   uint32 time = 1;  // time stamp at which the measurements were performed expressed in [ms]
```
CMurtagh-LGTM, line 225: Not needed here, people can read a proto file.

controllers/nugus_controller/nugus_controller.cpp
```diff
@@ -55,76 +81,317 @@ class NUgus : public webots::Robot {
...
+                }                    
+                CameraMeasurement measurement = *sensorMeasurements->add_cameras();
+                measurement.set_name(camera->getName());
+                measurement.set_width(camera->getWidth());
+                measurement.set_height(camera->getHeight());
+                measurement.set_quality(-1);  // raw image (JPEG compression not yet supported)
+                measurement.set_image((const char *)camera->getImage());
 
-                Nn = htonl(Nh);
+                // testing JPEG compression (impacts the performance)
```
CMurtagh-LGTM, line 263: Plz no commented code.

controllers/nugus_controller/nugus_controller.cpp
```diff
@@ -55,76 +81,317 @@ class NUgus : public webots::Robot {
...
+
+            // Parse message data
+            ActuatorRequests actuatorRequests;
+            if (!actuatorRequests.ParseFromArray(data.data(), Nh)) {
+                std::cerr << "Error: Failed to parse serialised message" << std::endl;
+                throw;
+            }
+
+            //------PARSE ACTUATOR REQUESTS MESSAGE-----------
+            // repeated MotorPosition motor_positions = 1;
```
CMurtagh-LGTM, line 137: Don't just copy the protobuf file.

controllers/nugus_controller/nugus_controller.cpp
```diff
@@ -15,38 +15,64 @@
...
+using controller::nugus::Force3DMeasurement;
+using controller::nugus::Vector3;
 
 class NUgus : public webots::Robot {
 public:
     NUgus(const int& time_step, const int& server_port)
         : time_step(time_step), server_port(server_port), tcp_fd(create_socket_server(server_port)) {
-            send(tcp_fd, "Welcome", 8, 0);
+        send(tcp_fd, "Welcome", 8, 0);
     }
```
CMurtagh-LGTM, line 70: A blocking constructor just feels wrong.

controllers/nugus_controller/nugus_controller.cpp
```diff
@@ -55,76 +81,317 @@ class NUgus : public webots::Robot {
...
 
-                // Parse message data
-                controller::nugus::RobotControl msg;
-                if (!msg.ParseFromArray(data.data(), Nh)) {
-                    std::cerr << "Error: Failed to parse serialised message" << std::endl;
-                    continue;
+                // repeated CameraQuality camera_qualities = 7;
+                // repeated CameraExposure camera_exposures = 8;
+                // Quality not implemented yet
+                // for (int i = 0; i < actuatorRequests.camera_qualities_size(); i++) {
```
CMurtagh-LGTM, line 156: No commented code plz

controllers/nugus_controller/nugus_controller.cpp
```diff
@@ -55,76 +81,317 @@ class NUgus : public webots::Robot {
...
+
+        if (send(tcp_fd, &Nn, sizeof(Nn), 0) < 0) {
+            std::cerr << "Error: Failed to send data over TCP connection: " << strerror(errno) << std::endl;
+        }
+        else if (send(tcp_fd, data.data(), data.size(), 0) < 0) {
+            std::cerr << "Error: Failed to send data over TCP connection: " << strerror(errno) << std::endl;
+        }
     }
 
+//     static void encode_jpeg(const unsigned char* image,
```
CMurtagh-LGTM, line 346: No commented code

CMurtagh-LGTM (changes requested):

controllers/nugus_controller/nugus_controller.cpp
ysims: You mean the send function being inside the constructor? Should it be removed?

controllers/nugus_controller/nugus_controller.cpp
ysims: I guess we can put it back in when they're ready with it.

controllers/nugus_controller/nugus_controller.cpp
CMurtagh-LGTM: The function `create_socket_server` blocks until a connection is formed.

Bidski: See #17

controllers/nugus_controller/nugus_controller.cpp
```diff
@@ -15,106 +15,322 @@
...
+        sensorMeasurements->set_real_time(real_time);
 
-                if (send(tcp_fd, &Nn, sizeof(Nn), 0) < 0) {
-                    std::cerr << "Error: Failed to send data over TCP connection: " << strerror(errno) << std::endl;
+        // Iterator over all the devices that have been enabled from any received ActuatorRequests messages
+        for (std::set<std::shared_ptr<webots::Device>>::iterator it = sensors.begin(); it != sensors.end(); ++it) {
+            // Try to cast to an accelerometer and if it works, add the accelerometer values to the SensorMeasurement
+            // message
+            std::shared_ptr<webots::Accelerometer> accelerometer =
+                std::dynamic_pointer_cast<webots::Accelerometer>(*it);
```
Bidski, line 229: This is not very likely to fail. All devices (accelerometers, etc) inherit from `webots::Device` and so casting it from a base class to a subclass will probably work.

Using [wb_get_node_type](https://cyberbotics.com/doc/reference/device?tab-language=c++#wb_device_get_node_type) on the device may be the better option

Bidski (changes requested):

controllers/nugus_controller/nugus_controller.cpp
```diff
@@ -15,107 +15,330 @@
...
+                    if (i < actuatorRequests.motor_forces_size()) {
+                        motor->setForce(actuatorRequests.motor_forces(i).force());
+                    }
+                    if (i < actuatorRequests.motor_torques_size()) {
+                        motor->setTorque(actuatorRequests.motor_torques(i).torque());
+                    }
+                    if (i < actuatorRequests.motor_pids_size()) {
+                        motor->setControlPID(motorPID.pid().x(), motorPID.pid().y(), motorPID.pid().z());
+                    }
                 }
```
Bidski, line 156: This is not the best way to handle this. We need to re-evaluate how we are going to handle these messages as it looks like forces/torques are an alternative control method to PIDs

[See this](https://cyberbotics.com/doc/reference/motor#wb_motor_set_force)
> As an alternative to the PID-controller, the wb_motor_set_force (resp. wb_motor_set_torque) function allows the user to directly specify the amount of force (resp. torque) that must be applied by a motor.

controllers/nugus_controller/nugus_controller.cpp
```diff
@@ -15,107 +15,330 @@
...
+        uint64_t real_time = tp.tv_sec * 1000 + tp.tv_usec / 1000;
+        sensorMeasurements->set_real_time(real_time);
 
-                if (send(tcp_fd, &Nn, sizeof(Nn), 0) < 0) {
-                    std::cerr << "Error: Failed to send data over TCP connection: " << strerror(errno) << std::endl;
+        // Iterator over all the devices that have been enabled from any received ActuatorRequests messages
+        for (std::set<webots::Device*>::iterator it = sensors.begin(); it != sensors.end(); ++it) {
+            // Try to cast to an accelerometer and if it works, add the accelerometer values to the SensorMeasurement
+            // message
+            webots::Accelerometer* accelerometer = reinterpret_cast<webots::Accelerometer*>(*it);
```
Bidski, line 237: I am not convinced that this is the correct way to handle this. I think looking at [getNodeType](https://cyberbotics.com/doc/reference/device?tab-language=c++#wb_device_get_node_type) would be the better way to decide what to cast to

Bidski (changes requested):

KipHamiltons:
@TrentHouliston do you think we should be using weak pointers instead of raw pointers for the Webots objects which we don't have ownership of?

Bidski:
> @TrentHouliston do you think we should be using weak pointers instead of raw pointers for the Webots objects which we don't have ownership of?

Do we know that the objects are managed by `shared_ptr`s?
> `std::weak_ptr` is a smart pointer that holds a non-owning ("weak") reference to an object that is managed by `std::shared_ptr`.

KipHamiltons:
> Do we know that the objects are managed by shared_ptrs?

Ahh, I misunderstood that... I think much of webots is written in C, so that seems unlikely

KipHamiltons:
I think that this PR should be merged, and that we can improve what exists such that we can run keyboardwalk in follow up PRs.

controllers/nugus_controller/nugus_controller.cpp
```diff
@@ -15,107 +15,452 @@
...
+                            break;
+                        }
+                    }
+                }
+                default:
+                    std::cerr << "Switch had no case. Unexpected WbNodeType: " << device->getNodeType() << std::endl;
+            }
+        }
+
+        if (false) { // set to true to print the created SensorMeasurements message for debugging
```
Bidski, line 340:
```suggestion
// Build in debug mode to print the created SensorMeasurements message for debugging
#ifndef NDEBUG
```

controllers/nugus_controller/nugus_controller.cpp
```diff
@@ -15,107 +15,452 @@
...
+                std::cout << "  sm.position_sensors: " << std::endl;
+                int i = 0;
+                for (auto sensor : sensorMeasurements->position_sensors()) {
+                    std::cout << "    sm.position_sensors[" << i << "]" << std::endl;
+                    std::cout << "      name: " << sensor.name() << std::endl;
+                    std::cout << "      value: " << sensor.value() << std::endl;
+                    i++;
                 }
             }
         }
```
Bidski, line 448:
```suggestion
        #endif // #NDEBUG
```

controllers/nugus_controller/nugus_controller.cpp
```diff
@@ -15,107 +15,452 @@
...
+        // Watch TCP file descriptor to see when it has input.
+        // No wait - polling as fast as possible
+        int num_ready = select(tcp_fd + 1, &rfds, nullptr, nullptr, &timeout);
+        if (num_ready < 0) {
+            std::cerr << "Error: Polling of TCP connection failed: " << strerror(errno) << std::endl;
+            throw;
+        }
+        else if (num_ready > 0) {
+            // Wire format
+            // unit32_t Nn  message size in bytes. The bytes are in network byte order (big endian)
```
Bidski, line 113: If `Nn` is zero, the likely corresponds to a protobuf message with all values set to their defaults

Bidski (changes requested):

controllers/nugus_controller/nugus_controller.cpp
ysims: I'm not sure what you mean, isn't this just handled by NUbots by not sending the message if we don't want to set it? Or do you mean we should have some sort of if else thing, ie don't let us set both (I'm not sure if that's what the docs are saying, that we can't set both?).

controllers/nugus_controller/nugus_controller.cpp
ysims: When you say defaults, do you mean empty? Is it the same thing as just leaving it as `ActuatorRequests actuatorRequests;`?

Bidski: It should be the same as that

KipHamiltons:
I reckon it'd be pretty cool if we could get this merged :wink: 
@Bidski @CMurtagh-LGTM

CMurtagh-LGTM (approved):
¯\\_(ツ)_/¯

PR #32, opened 2021-04-28
Updates camera parameters

ysims:
[Webots camera docs!](https://cyberbotics.com/doc/reference/camera)

[Taken from the robocup2021-model-april24 branch.](https://github.com/NUbots/NUWebots/tree/robocup2021-model-april24)


moosahassanx (commented):
With my little knowledge of Protobuf, I see nothing wrong with this.

protos/NUgus.proto
```diff
@@ -15,6 +15,12 @@ PROTO NUgus [
   field  SFBool      supervisor      FALSE              # Is `Robot.supervisor`.
   field  SFBool      synchronization TRUE               # Is `Robot.synchronization`.
   field  SFBool      selfCollision   FALSE              # Is `Robot.selfCollision`.
+  field SFFloat      fieldOfView          1.5707
+  field SFFloat      cameraWidth          640
+  field SFFloat      cameraHeight         480
```
JosephusPaye, line 20:
```suggestion
  field SFInt32      cameraWidth          640
  field SFInt32      cameraHeight         480
```

JosephusPaye (changes requested):

protos/NUgus.proto
```diff
@@ -5,17 +5,25 @@
...
+  field  MFString    robotTexture           "textures/NUgus_base_color.jpg"
+  field  SFString    name                   "NUgus"                           # Is `Robot.name`.
+  field  SFString    controller             "void"                            # Is `Robot.controller`.
+  field  MFString    controllerArgs         []                                # Is `Robot.controllerArgs`.
+  field  SFString    customData             ""                                # Is `Robot.customData`.
+  field  SFBool      supervisor             FALSE                             # Is `Robot.supervisor`.
+  field  SFBool      synchronization        TRUE                              # Is `Robot.synchronization`.
+  field  SFBool      selfCollision          FALSE                             # Is `Robot.selfCollision`.
+  # CAMERA PARAMETERS: See docs for more information
+  field SFFloat      fieldOfView            1.5707                            # Approximates PI/2 radians
```
JosephusPaye, line 19:
```suggestion
  field SFFloat      fieldOfView            1.5707                            # Approximates PI/2 radians (90 degrees)
```

protos/NUgus.proto
```diff
@@ -5,17 +5,25 @@
...
+  field SFInt32      cameraWidth            640                               # 640 pixels
+  field SFInt32      cameraHeight           480                               # 480 pixels
+  field SFFloat      cameraLensFocalLength  1.98                              # 1.98mm, as from the real cameras
```
JosephusPaye, line 22:
```suggestion
  field SFFloat      cameraLensFocalLength  1.98                              # 1.98mm, as from our real cameras (https://www.lensation.de/product/BF10M19828S118C/)
```

protos/NUgus.proto
```diff
@@ -1297,21 +1305,55 @@ PROTO NUgus [
...
+                          minFocalDistance 0.0   # 
+                        }
+                        # Allows the robot to detect objects in the world
+                        recognition DEF RECOGNITION Recognition {
+                          frameThickness 0       # Remove bounding boxes
+                          segmentation TRUE      # Add segmentation
+                        }
                       }
-                      Camera {
+                      DEF LEFT_CAMERA Camera {
```
Bidski, line 1340: Not sure we need the name capitalised?

protos/NUgus.proto
```diff
@@ -1297,21 +1305,55 @@ PROTO NUgus [
...
+                        
+                        # We are using a rectilinear lens because there is no true spherical lens
+                        spherical FALSE
+                        lens DEF LENS Lens {
+                          # Parameters are set to default
+                          center 0.5 0.5
+                          radialCoefficients 0 0
+                          tangentialCoefficients 0 0
+                        }
+                        focus DEF FOCUS Focus {
```
JosephusPaye, line 1328: @wongjoel and I were thinking of removing the Focus node for these reasons:

- It's directly related to FOV, and setting one should set the other. Setting both could possibly lead to conflicts/unwanted behavior? We don't know. Also, we're calculating the focal length dynamically based on the other parameters (in the vision data collection).
- Using a Focus node allows for changing the focus at run time, but our real lens have their focus pretty much fixed (e.g. they're preset based on how much the lens are screwed in).

I've tried running without the Focus node, and I couldn't see any difference. We can keep it though if its required for model inspection, and make sure we always set a value that doesn't conflict with the other related parameters.

protos/NUgus.proto
```diff
@@ -5,17 +5,25 @@
...
+  field  SFBool      supervisor             FALSE                             # Is `Robot.supervisor`.
+  field  SFBool      synchronization        TRUE                              # Is `Robot.synchronization`.
+  field  SFBool      selfCollision          FALSE                             # Is `Robot.selfCollision`.
+  # CAMERA PARAMETERS: See docs for more information
+  field SFFloat      fieldOfView            1.5707                            # Approximates PI/2 radians
+  field SFInt32      cameraWidth            640                               # 640 pixels
+  field SFInt32      cameraHeight           480                               # 480 pixels
+  field SFFloat      cameraLensFocalLength  1.98                              # 1.98mm, as from the real cameras
+  field SFFloat      cameraNoise            0.000000001                       # Not much noise on real cameras
+  field SFFloat      cameraMotionBlur       10                                # With over 100fps, real cameras do not have much motion blur
```
Bidski, line 24: do we have 100fps though?

Bidski (changes requested):

JosephusPaye (commented):
Some suggestions...

protos/NUgus.proto
ysims: This is just what I got from Trent, he said the real cameras run at over 100fps.

protos/NUgus.proto
ysims: Doesn't look like it's required for model inspection. If it's not there, it's just set to the default values and these are the default values, so there won't be any difference. This is the same for the lens parameters too.

protos/NUgus.proto
Bidski: Our real cameras do indeed run at around 170fps. I guess if we are going with realistic values this makes sense then

protos/NUgus.proto
```diff
@@ -1297,21 +1305,42 @@ PROTO NUgus [
...
+
+                        # Allows the robot to detect objects in the world
+                        recognition DEF recognition Recognition {
+                          frameThickness 0       # Remove bounding boxes
+                          segmentation TRUE      # Add segmentation
+                        }              
+                      }
+                      DEF right_camera Camera {
+                        name "right_camera"
+                       translation 0.0889361 -0.0352991 0.0713675
```
Bidski, line 1329:
```suggestion
                        translation 0.0889361 -0.0352991 0.0713675
```

Bidski (changes requested):

Bidski (approved):
Looks fine to me

JosephusPaye (approved):
LGTM

PR #34, opened 2021-05-15
[WIP]: Add scripted path when teleporting for visual mesh

moosahassanx:
Goal of PR:
1. Read modified config.yaml file (config_REFERENCE.yaml)
2. Modify/Redo "LOOP OVER ROTATIONS"
3. Make Python script to generate config.yaml files to be used when generating dataset pathways

TODO: 
- [ ] Have it compile for now.
- [ ] Verify that it's correct.


Lachourt:
Depending on what the end goal of this is, you might be better off making a new controller rather than merging into Ysobel's PR... See my latest PR #35 . I suppose this is avoided as long as the paths don't collide but having the vision data collection controller looking for a path config file might be conflicting when running the controller in #35

ysims:
Very out of date - if we do ever want this in the future, we can redo it. For now webots vision is fine.

PR #35, opened 2021-05-15
Add new controller that avoids collisions when teleporting

LachlanCourt:
The best way I found to avoid collisions was to have an entirely separate controller which actively moves all the other robots. Calculating the locations centrally means that we can check for collisions easily in one place. This involves adding a new generic Robot node and applying this controller to it, and all other robots can have a different controller applied to them simultaneously that can be grabbing vision data. Check out README to see how to set it up.
**EDIT** I just realised I forgot to run clang format - please forgive - I will do that when I take on board any other feedback


Lachourt:
r/RoastMe

controllers/teleport_master_controller/teleport_master_controller.cpp
```diff
@@ -0,0 +1,169 @@
...
+            newPos[0] = 4.3 - xDistrib(gen) / 100;
+            newPos[1] = 2.9 - yDistrib(gen) / 100;
+            newPos[2] = 0.51;
+            
+            // Loop through the vector of existing proposed locations and see if the new one is going to
+            // collide with any of them
+            for (std::array<double, 3> testPos : positions)
+            {
+              double distance = sqrt(pow((newPos[1] - testPos[1]), 2) + pow((newPos[0] - testPos[0]), 2));
+              if (distance < 0.1)
```
ysims, line 113: These hard coded values should probably be in a config file (or at the very least defined at the top).

controllers/teleport_master_controller/teleport_master_controller.cpp
```diff
@@ -0,0 +1,169 @@
...
+           // Convert the field to a vector to output to console
+           const double* target_translation_vec = target_translation_field.getSFVec3f();
+           
+           // Grab the current rotation field of the robot to modify
+           webots::Field& target_rotation_field = *(otherRobotsNodes[i])->getField("rotation");
+          // Convert the field to a vector to output to console
+          const double* target_rotation_vec = target_rotation_field.getSFRotation();
+          
+          // Output current location/rotation
+          std::ofstream log;
```
ysims, line 140: Use yaml to log?

controllers/teleport_master_controller/teleport_master_controller.cpp
```diff
@@ -0,0 +1,169 @@
...
+          }
+          // Loop until a proposed location has been found that doesn't clash with the existing ones
+          while (collision); 
+          // Finally add the proposed location in as a confirmed position
+          positions.push_back(newPos);
+        }
+
+
+        // Loop through every robot
+        for (long unsigned int i = 0; i < otherRobotsNodes.size(); i++)
```
ysims, line 127: `long`? Probs wont have that many robots on the field 😄

controllers/teleport_master_controller/teleport_master_controller.cpp
```diff
@@ -0,0 +1,169 @@
...
+        {
+          // Assume there is no collision
+          bool collision = false;
+          // newPos will be a "Proposed location" for a robot to teleport to
+          std::array<double, 3> newPos;
+          do
+          {
+            collision = false;
+            // Generate a new random location
+            newPos[0] = 4.3 - xDistrib(gen) / 100;
```
ysims, line 104: Would be good to get the field hard coded values from the actual field if possible.

ysims (changes requested):
If any of this, even just ideas, was from the Bit-Bots controller, make sure to credit them somewhere at the top of the file.

controllers/teleport_master_controller/teleport_master_controller.cpp
Lachourt: The 0.1 distance? I can move it up the top as a constant if you like. The rest are just the Euclidean distance formula

controllers/teleport_master_controller/teleport_master_controller.cpp
Lachourt: Do you want it as a list of like [time, location, rotation]?

controllers/teleport_master_controller/teleport_master_controller.cpp
Lachourt: The size of a vector comes in as a long unsigned int, it just comes up with a warning if I have a regular int so that was the easiest way to clear it. I could typecast the size to int if the concern is memory usage

controllers/teleport_master_controller/teleport_master_controller.cpp
Lachourt: I looked and looked around for this and had no luck, if you have any ideas that would be great but this was the only way I could figure out how to do it 😢 I can grab it from config too if that's cleaner but I couldn't work out how to do it dynamically

controllers/teleport_master_controller/teleport_master_controller.cpp
ysims: 0.1 distance, field values if we can't get them dynamically, robot height

controllers/teleport_master_controller/teleport_master_controller.cpp
ysims: Something like 
```yaml
# Simulation time
time: t
# Current robot position
location: [x,y,z]
# Angle axis rotation x,y,z,alpha
rotation: [x,y,z,alpha]
```

Lachourt: As an entry in a list though? Won't it overwrite with every new iteration of the loop if I just have them in the YAML as that?

ysims: Hmmm I guess then the question is why are we logging it? It might make more sense if each robot logs it's own information itself. If we want this information with the vision data information for eg localisation, then it should be happening when we get the images, ie in a different controller.

Lachourt: As long as those other controllers have a way to output to a different text file each - maybe have to pass in a filename as an argument... But yea just remove this bit then?

ysims: Yeah that's what the vision data collection controller does, since the visual mesh wants individual yaml files that correspond to image/s. Probs remove it yeah unless someone else thinks otherwise.

controllers/teleport_master_controller/README
```diff
@@ -0,0 +1,17 @@
+To use this controller, first assign a unique DEF name for each of the robots
+    that you want to control. Also set the supervisor field to TRUE.
+
+Next, add a new Robot Node to the webots world
+(Right click, Add New, Base Nodes, Robot)
+
+Set this new Robot arguments as follows
```
ysims, line 7: It'll probably be easier to just add such a node to the world file, ie add to the end of some world file the following:
```
DEF control Robot {
  controller "teleport_master_controller"
  controllerArgs [
    "control"
  ]
  supervisor TRUE
}
```

controllers/teleport_master_controller/teleport_master_controller.cpp
ysims: I've just done this, what you need to do is change the RobocupSoccerField node in the world file to have a def, ie
```
DEF FIELD RobocupSoccerField {
  size "kid"
}
```
Then you can get it using the supervisor from the def.
```
webots::Node* field = supervisor.getFromDef("FIELD");
```
And, for example, getting the name of the field and printing it out:
```
std::cout << field->getField("size")->getSFString() << std::endl;
```
There isn't a specific field for these values so you would need to change the proto so you can grab them.

Lachourt: Yeah okay, but you want it in the proto rather than the controller?
The thing is the range of random numbers I'm using isn't actually the size of the entire field - it's just the space inside the playing lines (Was getting weird errors when it was teleporting into goals by accident) is it still okay to have that in the proto?

ysims: I think it's ok to put in a field for it in the proto, just to expose it so we can grab the value. But then we'd set the actual value inside the proto (not in the field part, but in the node tree) using the dimensions (I believe thats what A,B,C etc are) to do some Lua code to add whatever is needed up.
Might be better too to use the full field but add the goals and ball in to the collision detection.

Lachourt: I can add the goals into the collision detection. I was told yesterday it probably won't be necessary but if you would like it I'll just need to grab them from def as well that's easy enough

controllers/teleport_master_controller/teleport_master_controller.cpp
KipHamiltons: It's a vector - is there a reason not to use a range-based `for`?

Lachourt: I need the index to apply the relevant position (Line 159)

KipHamiltons: oh true. can you use `size_t` to get rid of the warning here? It's more concise that `long unsigned int` :smile:

controllers/teleport_master_controller/teleport_master_controller.cpp
LachlanCourt: Okay so it appears that grabbing the goals to avoid collisions is going to be a major challenge... Because they're part of the "RobocupSoccerField" Node rather than objects in their own right, I am uncertain how to go about checking the actual goals. I can check for a collision but only if it has already moved there, which would still slow the simulation down.

controllers/teleport_master_controller/teleport_master_controller.cpp
```diff
@@ -0,0 +1,144 @@
...
+// https://github.com/bit-bots/wolfgang_robot/blob/feature/recognition/wolfgang_webots_sim/src/wolfgang_webots_sim/webots_camera_controller.py
+
+#include <array>
+#include <fstream>
+#include <iostream>
+#include <math.h>
+#include <random>
+#include <webots/Supervisor.hpp>
+
+#include "yaml-cpp/yaml.h"
```
Bidski, line 30: System includes should use angle brackets
```suggestion
#include <yaml-cpp/yaml.h>
```

controllers/teleport_master_controller/teleport_master_controller.cpp
```diff
@@ -0,0 +1,144 @@
...
+
+// AxisAngle rotations will be read from a config file and saved here
+std::vector<std::array<double, 4>> rotations;
+double minDistance;
+double zHeight;
+double xSize;
+double ySize;
```
Bidski, line 37: Is there a reason you made these globals?

controllers/teleport_master_controller/teleport_master_controller.cpp
```diff
@@ -0,0 +1,144 @@
...
+                // collide with any of them
+                for (std::array<double, 3> testPos : positions) {
+                    double distance = sqrt(pow((newPos[1] - testPos[1]), 2) + pow((newPos[0] - testPos[0]), 2));
+                    if (distance < minDistance) {
+                        collision = true;
+                    }
+                }
+            }
+            // Loop until a proposed location has been found that doesn't clash with the existing ones
+            while (collision);
```
Bidski, line 122:
```suggestion
            // Loop until a proposed location has been found that doesn't clash with the existing ones
            } while (collision);
```

controllers/teleport_master_controller/teleport_master_controller.cpp
```diff
@@ -0,0 +1,144 @@
...
+            // Finally add the proposed location in as a confirmed position
+            positions.push_back(newPos);
+        }
+
+        // Loop through every robot
+        for (size_t i = 0; i < otherRobotsNodes.size(); i++) {
+            // Grab translation field of the robot to modify
+            webots::Field& target_translation_field = *(otherRobotsNodes[i])->getField("translation");
+            // Grab the current rotation field of the robot to modify
+            webots::Field& target_rotation_field = *(otherRobotsNodes[i])->getField("rotation");
```
Bidski, line 132: Just leave these as pointers. There is no real need/benefit to dereferencing them
```suggestion
            webots::Field* target_translation_field = otherRobotsNodes[i]->getField("translation");
            // Grab the current rotation field of the robot to modify
            webots::Field* target_rotation_field = otherRobotsNodes[i]->getField("rotation");
```

controllers/teleport_master_controller/teleport_master_controller.cpp
```diff
@@ -0,0 +1,144 @@
...
+
+    // Load def arguments of other robots in the field into a vector
+    std::vector<webots::Node*> otherRobotsNodes;
+    // Only one supervisor instance can exist in a single controller. As such the same one will be used
+    // every time
+    webots::Supervisor supervisor = webots::Supervisor();
+    for (int i = 2; i < argc; i++) {
+        // Load the Node of the robot by def argument and add to vector
+        webots::Node* tempNode = supervisor.getFromDef(argv[i]);
+        otherRobotsNodes.push_back(tempNode);
```
Bidski, line 59:
```suggestion
        otherRobotsNodes.emplace_back(supervisor.getFromDef(argv[i]));
```

controllers/teleport_master_controller/teleport_master_controller.cpp
```diff
@@ -0,0 +1,144 @@
...
+        return 1;
+    }
+    catch (const YAML::ParserException& e) {
+        std::cerr << e.msg << std::endl;
+        return 2;
+    }
+
+    webots::Node* tempNode = supervisor.getFromDef("test");
+    xSize                  = tempNode->getField("xSize")->getSFFloat();
+    ySize                  = tempNode->getField("ySize")->getSFFloat();
```
Bidski, line 80:
```suggestion
    const double xSize                  = tempNode->getField("xSize")->getSFFloat();
    const double ySize                  = tempNode->getField("ySize")->getSFFloat();
```

controllers/teleport_master_controller/teleport_master_controller.cpp
```diff
@@ -0,0 +1,144 @@
...
+    int timeStep = int(supervisor.getBasicTimeStep());
+
+    // Generate random seed
+    std::random_device rd;   // Will be used to obtain a seed for the random number engine
+    std::mt19937 gen(rd());  // Standard mersenne_twister_engine seeded with rd()
+
+    // Generate distributions for random number systems in the following loop
+    std::uniform_int_distribution<> xDistrib(0, xSize * 100);
+    std::uniform_int_distribution<> yDistrib(0, ySize * 100);
+    std::uniform_int_distribution<> rotDistrib(0, rotations.size() - 1);
```
Bidski, line 92:
```suggestion
    std::uniform_real_distribution<> xDistrib(-xSize * 0.5, xSize * 0.5);
    std::uniform_real_distribution<> yDistrib(-ySize * 0.5, ySize * 0.5);
    std::uniform_real_distribution<> rotDistrib(0.0, rotations.size() - 1.0);
```

controllers/teleport_master_controller/teleport_master_controller.cpp
```diff
@@ -0,0 +1,144 @@
...
+        for (size_t i = 0; i < otherRobotsNodes.size(); i++) {
+            // Assume there is no collision
+            bool collision = false;
+            // newPos will be a "Proposed location" for a robot to teleport to
+            std::array<double, 3> newPos;
+            do {
+                collision = false;
+                // Generate a new random location
+                newPos[0] = xSize / 2 - xDistrib(gen) / 100;
+                newPos[1] = ySize / 2 - yDistrib(gen) / 100;
```
Bidski, line 109:
```suggestion
                newPos[0] = xDistrib(gen);
                newPos[1] = yDistrib(gen);
```

controllers/teleport_master_controller/teleport_master_controller.cpp
```diff
@@ -0,0 +1,144 @@
...
+                newPos[2] = zHeight;
+
+                // Loop through the vector of existing proposed locations and see if the new one is going to
+                // collide with any of them
+                for (std::array<double, 3> testPos : positions) {
+                    double distance = sqrt(pow((newPos[1] - testPos[1]), 2) + pow((newPos[0] - testPos[0]), 2));
```
Bidski, line 115:
```suggestion
                    double distance = std::sqrt(std::pow((newPos[1] - testPos[1]), 2) + std::pow((newPos[0] - testPos[0]), 2));
```

controllers/teleport_master_controller/teleport_master_controller.cpp
```diff
@@ -0,0 +1,144 @@
...
+ * Copyright 2021 NUbots <nubots@nubots.net>
+ */
+ 
+// Special thanks to Hamburg Bit-Bots for ideas for this controller
+// https://github.com/bit-bots/wolfgang_robot/blob/feature/recognition/wolfgang_webots_sim/src/wolfgang_webots_sim/webots_camera_controller.py
+
+#include <array>
+#include <fstream>
+#include <iostream>
+#include <math.h>
```
Bidski, line 26:
```suggestion
#include <cmath>
```

Bidski (changes requested):

controllers/teleport_master_controller/teleport_master_controller.cpp
LachlanCourt: Not really, I can pull them inside main

controllers/teleport_master_controller/teleport_master_controller.cpp
LachlanCourt: What's the difference between emplace and push?

controllers/teleport_master_controller/teleport_master_controller.cpp
LachlanCourt: Oh that's much cleaner I didn't realise I could do that

controllers/teleport_master_controller/teleport_master_controller.cpp
LachlanCourt: What's the benefit of using this library over math.h?

controllers/teleport_master_controller/teleport_master_controller.cpp
LachlanCourt: What's the purpose of specifying std here?

tayloryoung6396: Because it's part of the std library?

controllers/teleport_master_controller/teleport_master_controller.cpp
tayloryoung6396: Destroy all global variables! 
image

controllers/teleport_master_controller/teleport_master_controller.cpp
TrentHouliston: You should always be prefixing your standard library functions with `std::` when you're doing them in c++

In this case in particular it's even more important as if you don't it uses the C version of the functions which will automatically promote the input variables to doubles as C doesn't have the concept of overloading.
If you use the `std::` versions it will use the type of input that you are using (so if you put in floats it will use the float version). Since in this case, you are already using doubles it probably doesn't make any practical difference but it's a habit you should be getting into anyway.

controllers/teleport_master_controller/teleport_master_controller.cpp
TrentHouliston: `math.h` is the C header, `<cmath>` is the c++ header and we are using c++

Technically according to the c++ standard, you would have been expected to wrap your `<math.h>` in an extern "C" block but basically every standard library does that for you in macros.

controllers/teleport_master_controller/teleport_master_controller.cpp
TrentHouliston: A list of elements has memory allocated to it. When you use push what it does is it makes new memory, constructs an instance of your object there, then copies that memory into the memory in the list.

Emplace instead takes the arguments to the constructor and constructs the new object directly in the memory that was already there for the list.

So basically emplace means one less copy than push

controllers/teleport_master_controller/teleport_master_controller.cpp
TrentHouliston: I mean, I do this all the time so this one is a matter of opinion :P

It means I don't have to use `->` later!

controllers/teleport_master_controller/teleport_master_controller.cpp
TrentHouliston: Globals are bad and should be avoided at almost all costs. I'm not sure where you should be storing this state but in general globals are just asking for trouble

Bidski: Since this controller only consists of a single main function there is no reason not to just put them in main, and no benefit gained by making them global

controllers/teleport_master_controller/teleport_master_controller.cpp
LachlanCourt: I originally changed them from pointers I think because we were trying to avoid using raw pointers as much as possible and I couldn't get smart pointers to work with webots... But I don't mind either way :)

controllers/teleport_master_controller/teleport_master_controller.cpp
TrentHouliston: There's never a benefit to be gained by making variables global! Only future pain :'(

controllers/teleport_master_controller/teleport_master_controller.cpp
LachlanCourt: So it's a case of making it run with less strain on memory resources? Why does push even exist in that case - is there a benefit?

controllers/teleport_master_controller/teleport_master_controller.cpp
Bidski: The only other potential issue are other libraries defining these functions in the global namespace. We aren't using any libraries that do this horrendous practice, but as @TrentHouliston said, its a good habit to get into

LachlanCourt: Gotcha. The weird back and forth between C++ using some C libraries still confuses me sometimes

controllers/teleport_master_controller/teleport_master_controller.cpp
```diff
@@ -0,0 +1,139 @@
...
+            do {
+                collision = false;
+                // Generate a new random location
+                newPos[0] = xDistrib(gen);
+                newPos[1] = yDistrib(gen);
+                newPos[2] = zHeight;
+
+                // Loop through the vector of existing proposed locations and see if the new one is going to
+                // collide with any of them
+                for (std::array<double, 3> testPos : positions) {
```
Bidski, line 109: `const` because you aren't modifying the arrays. `auto` because less characters (although this is optional, main benefit IMO is when you either don't know (or care about) the type or when the type is just template messiness). `&` to avoid making copies of the arrays
```suggestion
                for (const auto& testPos : positions) {
```

controllers/teleport_master_controller/teleport_master_controller.cpp
```diff
@@ -0,0 +1,139 @@
...
+                    double distance =
```
Bidski, line 110:
```suggestion
                    const double distance =
```

controllers/teleport_master_controller/teleport_master_controller.cpp
LachlanCourt: I see I see do they not provide the same functionality?

controllers/teleport_master_controller/teleport_master_controller.cpp
```diff
@@ -0,0 +1,139 @@
...
+    // Generate distributions for random number systems in the following loop
+    std::uniform_real_distribution<> xDistrib(-xSize * 0.5, xSize * 0.5);
+    std::uniform_real_distribution<> yDistrib(-ySize * 0.5, ySize * 0.5);
+    std::uniform_real_distribution<> rotDistrib(0.0, rotations.size() - 1.0);
+
+    while (supervisor.step(timeStep) != -1) {
+
+        // Declare a vector of positions that will be saved as they are randomly generated, to be later
+        // applied to each robot
+        std::vector<std::array<double, 3>> positions = std::vector<std::array<double, 3>>();
```
Bidski, line 93: No need to do this. You are effectively calling the default constructor (twice, I think) and then copying that into `positions`. Without the extra bit you are just calling the default constructor.
```suggestion
        std::vector<std::array<double, 3>> positions;
```

controllers/teleport_master_controller/teleport_master_controller.cpp
```diff
@@ -0,0 +1,139 @@
...
+                    double distance =
+                        std::sqrt(std::pow((newPos[1] - testPos[1]), 2) + std::pow((newPos[0] - testPos[0]), 2));
+                    if (distance < minDistance) {
+                        collision = true;
+                    }
+                }
+                // Loop until a proposed location has been found that doesn't clash with the existing ones
+            } while (collision);
+            // Finally add the proposed location in as a confirmed position
+            positions.push_back(newPos);
```
Bidski, line 119:
```suggestion
            positions.emplace_back(newPos);
```

Bidski (changes requested):

controllers/teleport_master_controller/teleport_master_controller.cpp
Bidski: For the most part, yes. The C++ library does more though. It provides overloading so the same function can be used with different types (C can't do this at all, instead they have something like `sqrt` for double and `sqrtf` for floats)
```c++
float sqrt(float f);
double sqrt(double f);
```

Probably does more, but most if it is around overloading and better type support

controllers/teleport_master_controller/teleport_master_controller.cpp
Bidski: Not so much strain on memory resources, for the most part (and in most circumstances) you are probably talking about a handful of bytes, although this could be important in some circumstances. There are extra computational resources involved in the copying though which is probably more costly.

There will be benefits to using push in some circumstances. I am currently too tired to actually think of a case though. You got any ideas @TrentHouliston  ?

controllers/teleport_master_controller/teleport_master_controller.cpp
LachlanCourt: Oh tright I didn't realise that was what that did. I was treating it like a primitive array where you gotta initialise it first

controllers/teleport_master_controller/teleport_master_controller.cpp
LachlanCourt: That is definitely much cleaner

controllers/teleport_master_controller/teleport_master_controller.cpp
Bidski: As a simple rule of thumb (this only applies to the C and C++ standard library header files).

If the header file that you are including has a `.h` on the end of it, it is a C standard library header file (e.g. `math.h` and `stdio.h`).

If the header file has no extension at all, it is a C++ standard library header file (e.g. `cmath` and `iostream`).

If you want to use a C standard library header file in a C++ program (e.g. you want the math functions), generally you drop the extension and put a `c` on the from (e.g. `math.h` becomes `cmath`, `stdlib.h` becomes `cstdlib`). Always worth looking at [en.cppreference.com](https://en.cppreference.com/w/cpp) to see what the correct header file to include is

controllers/teleport_master_controller/teleport_master_controller.cpp
LachlanCourt: Yeah okay gotcha. Thanks :)

TrentHouliston: Yeah if you include "math.h" it often doesn't include the std:: versions of the functions (although sometimes it does).
This is where keeping to the standard rather than "what works" is important in c++.
If you just go with whatever works you can end up trying to compile your code on another compiler (clang vs gcc vs msvc) and it'll sometimes work and sometimes not which can drive you insane.

If you stick to the standard then it'll work everywhere (unless there is a bug in the compiler :P)

controllers/teleport_master_controller/teleport_master_controller.cpp
```diff
@@ -0,0 +1,139 @@
...
+        for (size_t i = 0; i < otherRobotsNodes.size(); i++) {
+            // Grab translation field of the robot to modify
+            webots::Field* target_translation_field = otherRobotsNodes[i]->getField("translation");
+            // Grab the current rotation field of the robot to modify
+            webots::Field* target_rotation_field = otherRobotsNodes[i]->getField("rotation");
+
+            // There will be a position for every robot in the positions vector
+            target_translation_field->setSFVec3f(positions[i].data());
+            // Apply new rotation
+            target_rotation_field->setSFRotation(rotations[rotDistrib(gen)].data());
```
Bidski, line 132: I just noticed that you only use each of these variables once. I would just do this
```suggestion
            // There will be a position for every robot in the positions vector
            otherRobotsNodes[i]->getField("translation")->setSFVec3f(positions[i].data());
            // Apply new rotation
            otherRobotsNodes[i]->getField("rotation")->setSFRotation(rotations[rotDistrib(gen)].data());
```

Bidski (changes requested):

controllers/teleport_master_controller/teleport_master_controller.cpp
LachlanCourt: I was considering that, I decided to leave it split up for readability but I can merge them if that's more suitable

controllers/teleport_master_controller/teleport_master_controller.cpp
TrentHouliston: There are very seldom cases where you should use push rather than emplace, mostly to do with weird cases where things don't have the right constructors/operators.

The history of this is `push_back` is the original and has been around since the dinosaurs roamed the earth. `emplace_back` was added in c++11 and is better in almost every way.

controllers/teleport_master_controller/teleport_master_controller.cpp
LachlanCourt: Amazing, thanks for that!

controllers/teleport_master_controller/teleport_master_controller.cpp
TrentHouliston: If it was a primitive array that you had to initialise first this is still the wrong way of doing it :P
Also don't use primitive arrays in c++ nowadays, use `std::array<type, count>` and initialise it by doing
```cpp
std::array<int, 5> thing{};
```
Which will set it to all 0s (or all default constructed elements if they are not primitives)

controllers/teleport_master_controller/teleport_master_controller.cpp
TrentHouliston: Once you're no longer dealing with standard library headers though all bets are off...

controllers/teleport_master_controller/teleport_master_controller.cpp
LachlanCourt: Oh my mind was in java town when I said that. But yea Kip got me using the std::array last time, it's much nicer :D

controllers/teleport_master_controller/teleport_master_controller.cpp
LachlanCourt: Yeah okay neat so pretty much use emplace_back unless it doesn't work and then try push_back :rofl:

controllers/teleport_master_controller/teleport_master_controller.cpp
JosephusPaye: The immediate convenience is certainly a gain 😛

Bidski (commented):
Well the code looks OK to me. Haven't tested it though, so someone else who has tested it will need to approve it

LachlanCourt:
Wonderful, thanks Alex :)

controllers/teleport_master_controller/teleport_master_controller.cpp
ysims: If it works as is currently then I don't think we should worry about it too much. You'd need to deal with transformations to do it as a child of the field, and it's probably fine as is.

controllers/teleport_master_controller/README.txt
```diff
@@ -0,0 +1,19 @@
+To use this controller:
+
+First add the following lines to the "kid.wbt" world file
```
ysims, line 3:
```suggestion
First add the following lines to the world file
```
Isn't necessarily the kid world file.

ysims (approved):
LGTM

LachlanCourt:
:tada:

PR #36, opened 2021-05-15
Adds ASan and UBSan as build options

KipHamiltons:
Address sanitizer and undefined behaviour sanitizer can be useful in finding and debugging buggy code.
This PR adds options to the cmake configuration to enable them.
I have tested these by using `ldd` on `nugus_controller`, and `ldd` reports that `nugus_controller` is linked to the sanitizers if they are enabled.


Bidski (approved):
Seems OK to me. Can you enable both ASan and UBSan at the same time? Do they play well together?

KipHamiltons:
I can't see anything online suggesting that they don't. I just tested building with them both on and it compiled fine. I ran the controller with webots and it didn't crash or anything, so I suppose so?

PR #37, opened 2021-05-16
Changes suggested by clang-tidy

KipHamiltons:
Follow up PR adds clang-tidy to the repo

PR #38, opened 2021-05-16
Add clang tidy

KipHamiltons:
This PR adds clang-tidy to the build system. 
I am not sure if we'll need to add a dependency for it to the repo for people to get, so I await other people's tests to confirm we don't.
This also adds a capability to automatically apply fixes after building your code with `make apply-clang-tidy`. I am not sure where we should document that, but it's worth documenting IMO. 

This PR should wait for the fixes which it suggests: #37


controllers/nugus_controller/CMakeLists.txt
```diff
@@ -20,25 +20,34 @@ add_custom_command(
...
+
+# Create an executable using all of the sources and the non-clang-tidy suitable generated protobuf sources
+add_executable(nugus_controller ${controller_sources} $<TARGET_OBJECTS:robot_control_protobuf>)
 
 # Make sure we can find the utility folder
 target_include_directories(nugus_controller PRIVATE "${PROJECT_SOURCE_DIR}/shared")
 
-# Make sure we can find the generated protobuf header files
-target_include_directories(nugus_controller PRIVATE "${CMAKE_CURRENT_BINARY_DIR}/protobuf")
+# # Make sure we can find the generated protobuf header files
```
Bidski, line 49:
```suggestion
# Make sure we can find the generated protobuf header files
```

Bidski (changes requested):

KipHamiltons:
There are a few warnings, since I didn't merge this fast enough :grimacing:
I might make a follow up fixing those I suppose :(

PR #39, opened 2021-05-16
Adds cut-down b script

KipHamiltons:
Having the same interface for configuring, building, and formatting our code as the main codebase is nice. This is the foundation of the `b` interface, for this repo

### DIFF
--- a/scripts/b.py
+++ b/scripts/b.py
@@ -1,0 +1,136 @@
+#!/usr/bin/env python3
+import argparse
+import os
+import pkgutil
+import re
+import sys
+
+# Don't make .pyc files
+sys.dont_write_bytecode = True
+
+scripts_dir = os.path.dirname(os.path.realpath(__file__))
+project_dir = os.path.dirname(scripts_dir)
+build_dir = os.path.join(project_dir, "build")
+
+# Build our cmake cache
+cmake_cache = {}
+
+dirs = [build_dir]
+try:
+    dirs.extend([os.path.join(build_dir, f) for f in os.listdir(build_dir)])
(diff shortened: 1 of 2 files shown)


scripts/b.py
```diff
@@ -0,0 +1,136 @@
...
+cmake_cache = {}
+
+dirs = [build_dir]
+try:
+    dirs.extend([os.path.join(build_dir, f) for f in os.listdir(build_dir)])
+except FileNotFoundError:
+    pass
+
+for d in dirs:
+    if os.path.isfile(os.path.join(project_dir, d, "CMakeCache.txt")):
```
Bidski, line 25: `d` should already be an absolute path
```suggestion
    if os.path.isfile(os.path.join(d, "CMakeCache.txt")):
```

scripts/b.py
```diff
@@ -0,0 +1,136 @@
...
+                        module.run(**vars(command.parse_args()))
+
+                        # We're done, exit
+                        exit(0)
+
+    # If we reach this point, we couldn't find a tool to use.
+    # In this case we need to look through all the tools so we can register them all.
+    # This will provide a complete help for the function call so the user can try again
+    tools = {}
+    for importer, modname, ispkg in pkgutil.walk_packages([user_tools_path, nuclear_tools_path]):
```
Bidski, line 105: These variable don't exist. I assume we want this?
```suggestion
    for importer, modname, ispkg in pkgutil.walk_packages([scripts_dir]):
```

Bidski (changes requested):

Bidski (approved):
LGTM

PR #41, opened 2021-05-17
Adds formatting script

KipHamiltons:
`./b format` has arrived.
This PR also makes the formatting fixes which is would do when run on the current codebase.

Depends on #39

### DIFF
--- a/scripts/format.py
+++ b/scripts/format.py
@@ -17,7 +17,7 @@ def _do_format(path):
         # Use the absolute path to file
         abs_path = os.path.abspath(path)
 
-        if path.endswith((".h", ".c", ".cc", ".cxx", ".cpp", ".hpp", ".ipp")):
+        if path.endswith((".h", ".c", ".cc", ".cxx", ".cpp", ".hpp", ".ipp", ".proto")) and "protos" not in path:
             text = "Formatting {} with clang-format\n".format(path)
             text = text + check_output(["clang-format", "-i", "-style=file", abs_path], stderr=STDOUT).decode("utf-8")
         elif path.endswith(("CMakeLists.txt", ".cmake")):
--- a/scripts/format.py
+++ b/scripts/format.py
@@ -37,10 +37,9 @@ def _do_format(path):
 
 def register(command):
     command.help = "Format all the code in the codebase using clang-format"
-    print("nice kevin")
+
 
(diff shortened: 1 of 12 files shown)


scripts/format.py
```diff
@@ -0,0 +1,54 @@
...
+        text = text + e.output.decode("utf-8")
+    except:
+        cprint("You don't have one of the formatters installed.\nRun `sudo -H pip3 install -r requirements.txt`", color='red')
+
+    return text
+
+
+def register(command):
+    command.help = "Format all the code in the codebase using clang-format"
+    print("nice kevin")
```
Bidski, line 40: Leave kevin alone!

scripts/format.py
```diff
@@ -0,0 +1,54 @@
...
+
+def run(**kwargs):
+    print("nice one!!")
```
Bidski, line 43: Is it though?

scripts/format.py
```diff
@@ -0,0 +1,54 @@
...
+import b
+
+
+def _do_format(path):
+    text = ""
+    try:
+        # Use the absolute path to file
+        abs_path = os.path.abspath(path)
+
+        if path.endswith((".h", ".c", ".cc", ".cxx", ".cpp", ".hpp", ".ipp")):
```
Bidski, line 20: We should look for `proto` files that are not in the `protos` folder and format them as well

Bidski (changes requested):
You are missing both `.clang-format` and `.cmake-format.py` files

I didn't review `b` and `scripts/b.py` here as they are in #39

scripts/format.py
JosephusPaye: Is this a self reference to _Big Kevin_? 😅

KipHamiltons: hahhahaaha anyone who can think of something that doesn't sound stupid for debug print statements is a psychopath

scripts/format.py
KipHamiltons: :joy: :joy:

KipHamiltons:
We already have a `.clang-format` and `.cmake-format.py` file in this repo, although the `.clang-format` was slightly out of date, so it had been brought up to parity with the NUbots one.

Bidski (approved):
LGTM

PR #43, opened 2021-05-18
Use a poll-based method for accepting TCP connections

Bidski:
Addresses issue #17

### DIFF
--- a/shared/utility/tcp.hpp
+++ b/shared/utility/tcp.hpp
@@ -30,9 +30,8 @@
     #include <netdb.h>      /* definition of gethostbyname */
     #include <netinet/in.h> /* definition of struct sockaddr_in */
     #include <poll.h>       /* definition of poll and pollfd */
-    #include <sys/socket.h>
-    #include <sys/time.h>
-    #include <unistd.h> /* definition of close */
+    #include <sys/socket.h> /* definition of socket, accept, listen, and bind */
+    #include <unistd.h>     /* definition of close */
 #endif
 
 namespace utility::tcp {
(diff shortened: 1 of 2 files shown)


Bidski:
@TrentHouliston and @CMurtagh-LGTM and @KipHamiltons this is currently untested. Just getting feedback before I go further

shared/utility/tcp.hpp
```diff
@@ -73,39 +73,72 @@ inline int create_socket_server(const int& port) {
...
+            std::cerr << "Error: Cannot accept client connection on port " << port << ": " << strerror(errno)
+                      << std::endl;
+            close_socket(server_fd);
+            return -1;
+        }
+
+        // Get client information
         const hostent* client_info = gethostbyname(inet_ntoa(client.sin_addr));
-        std::cerr << "Accepted connection from: " << client_info->h_name << std::endl;
+        std::cerr << "Accepted connection on port " << port << " from: " << client_info->h_name << std::endl;
```
CMurtagh-LGTM, line 132: This should be a `cout` instead of a `cerr` as it is not an error.

CMurtagh-LGTM:
Looks good

shared/utility/tcp.hpp
KipHamiltons: It's not an error, but it's not a detail that you would want to know unless you're debugging/monitoring, which is what the error stream is for, no?

Bidski: That is basically the logic I was going for, but happy to change it

CMurtagh-LGTM: @KipHamiltons We're running this in Webots, basically the only difference that realistically happens is red text vs black text.
We should reserve the `cerr` for errors as it makes it simpler to see when there is erroneous vs expected behavior.

Bidski: Works for me

KipHamiltons: Good point that this is a webots stream. Fair enough

KipHamiltons (commented):
We shouldn't forget about this PR. The code looks fine, but whether it works is obviously the important thing

PR #46, opened 2021-05-20
Add more compiler warnings and function for setting them

KipHamiltons:
I found a really nice comprehensive list of compiler warnings and a script/function to add them.
This adds them for this codebase.

should I do something similar for NUbots?


cmake/CompilerWarnings.cmake
```diff
@@ -0,0 +1,48 @@
...
+      -Wnon-virtual-dtor # warn the user if a class with virtual functions has a non-virtual destructor. This helps
+                         # catch hard to track down memory errors
+      -Wold-style-cast # warn for c-style casts
+      -Wcast-align # warn for potential performance problem casts
+      -Wunused # warn on anything being unused
+      -Woverloaded-virtual # warn if you overload (not override) a virtual function
+      -Wpedantic # warn if non-standard C++ is used
+      -Wconversion # warn on type conversions that may lose data
+      -Wsign-conversion # warn on sign conversions
+      -Wnull-dereference # warn if a null dereference is detected
```
Bidski, line 20: Only in GCC >= 6.0

CMakeLists.txt
```diff
@@ -12,6 +12,12 @@ if(NOT CMAKE_BUILD_TYPE AND NOT CMAKE_CONFIGURATION_TYPES)
...
 endif()
 
+# We need to tell cmake where the function for the compiler warnings is
+list(APPEND CMAKE_MODULE_PATH ${CMAKE_SOURCE_DIR}/cmake)
+include(CompilerWarnings)
+
+set_project_warnings()
+
 # We use additional modules that cmake needs to know about
 list(APPEND CMAKE_MODULE_PATH "${PROJECT_SOURCE_DIR}/cmake/Modules/")
```
Bidski, line 22: Lines 15 and 22 could be combined

Bidski: `option(WARNINGS_AS_ERRORS "Treat compiler warnings as errors" TRUE)` should be added somewhere as well

Bidski (changes requested):
Seems like a good idea. We probably do want to add this for NUbots too

KipHamiltons:
this introduces just one compiler warning from protobuf, and I can't work out how to remove it haha
```
[1/3] Building CXX object controllers/nugus_controller/CMakeFiles/robot_control_protobuf.dir/protobuf/RobotControl.pb.cc.o
controllers/nugus_controller/protobuf/RobotControl.pb.cc: In member function ‘virtual const char* controller::nugus::RobotControl::_InternalParse(const char*, google::protobuf::internal::ParseContext*’:
controllers/nugus_controller/protobuf/RobotControl.pb.cc:98:65: warning: conversion from ‘google::protobuf::uint64’ {aka ‘long unsigned int’} to ‘google::protobuf::int32’ {aka ‘int’} may change value -Wconversion]
   98 |           num_ = ::PROTOBUF_NAMESPACE_ID::internal::ReadVarint64(&ptr);
```

TrentHouliston:
The way I delt with all the warnings in the generated code is I didn't have any warnings when compiling those files (I made them a separate library)

Bidski:
> this introduces just one compiler warning from protobuf, and I can't work out how to remove it haha
> ```
> [1/3] Building CXX object controllers/nugus_controller/CMakeFiles/robot_control_protobuf.dir/protobuf/RobotControl.pb.cc.o
> controllers/nugus_controller/protobuf/RobotControl.pb.cc: In member function ‘virtual const char* controller::nugus::RobotControl::_InternalParse(const char*, google::protobuf::internal::ParseContext*’:
> controllers/nugus_controller/protobuf/RobotControl.pb.cc:98:65: warning: conversion from ‘google::protobuf::uint64’ {aka ‘long unsigned int’} to ‘google::protobuf::int32’ {aka ‘int’} may change value -Wconversion]
>    98 |           num_ = ::PROTOBUF_NAMESPACE_ID::internal::ReadVarint64(&ptr);
> ```

I had to disable a number of warnings on the generated protobuf files in the nubots codebase (probably in nuclear roles neutrons)

TrentHouliston:
Yeah, one of the things that I did in the cmake modernise stuff is to move protobufs into their own nonsense land with no warnings so that NUbots doesn't disable all those warnings in the actual codebase

cmake/CompilerWarnings.cmake
```diff
@@ -0,0 +1,43 @@
...
+
+  if(CMAKE_CXX_COMPILER_ID MATCHES ".*Clang")
+    set(PROJECT_WARNINGS ${CLANG_WARNINGS})
+  elseif(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
+    set(PROJECT_WARNINGS ${GCC_WARNINGS})
+  else()
+    message(AUTHOR_WARNING "No compiler warnings set for '${CMAKE_CXX_COMPILER_ID}' compiler.")
+  endif()
+
+  add_compile_options(${PROJECT_WARNINGS})
```
Bidski, line 41:
```suggestion
  if(NOT ${ENABLE_CLANG_TIDY})
    add_compile_options(${PROJECT_WARNINGS})
  else()
    message(STATUS "Static analyser enabled. Not applying any extra compiler warnings")
  endif()
```

Bidski (changes requested):

cmake/CompilerWarnings.cmake
KipHamiltons: We want the extra warnings. Clang-tidy doesn't all of them enabled, and we removed the ones which made clang-tidy fail to run, because clang-tidy did have checks for those things.

Bidski: Was it only the extra GCC warnings that caused clang to fail? If so, add those back in and then only set the clang warnings if we are using clang-tidy.

Bidski (approved):
LGTM

PR #47, opened 2021-05-20
Add CI pipeline

KipHamiltons:
Depends on #38 and #46


Y0sh1dk:
What needs to be changed on our buildkite agents for them to start running this pipeline?

KipHamiltons:
AFAIK the agents set up for the NUbots organisation will automatically run these once I have set up the pipeline in buildkite

TrentHouliston:
Keep the CI_BUILD flag, it's there so that all your CI flags can be done in the CMake file so if you need to test a CI build locally it's easy. CI_BUILD=On should be your only flag needed in the CI configuration step

KipHamiltons:
Here's my line of thinking.
Now that we're going to enable heaps more warnings, we have a set of warnings which can't be enabled when running clang-tidy, because clang doesn't have those gcc warnings and it errors if you try to run it with them.
This means that to run clang-tidy, we can't use all the warnings.
This means we're going to have two steps in the pipeline, to get full warning coverage - one with clang-tidy and one with the full set of gcc warnings.
This means that the only thing that the `CI_BUILD` was going to do was enable `WARNINGS_AS_ERRORS`, because if it enabled clang-tidy, we don't get full warning coverage.

So basically all `CI_BUILD` did was set `WARNINGS_AS_ERRORS`. That seems overly pointless to me.

Doing a full `CI_BUILD` dry-run locally is going to be painful anyway, because of these extra warnings not working with clang-tidy. It's going to be two steps. If you have a suggestion which makes it one step that would be greatly appreciated - although removing the extra gcc warnings seems like a lame option, and that seems to be the requirement.
I suppose we could have a CI dry-run tool, which checks everything. that would probably be good.

KipHamiltons:
I have to install the dependencies of this repo onto the buildkite agents, but once I have done that, this should be good to go

KipHamiltons:
We have a successful build :tada: 
Still finalising webots installation on all of the agents though, but we're very close to being all systems go :rocket:

Bidski (changes requested):
When #46 is merged we will need to update the if statement in `cmake/CompilerWarnings.cmake` to also check for `CI_BUILD`

Bidski (approved):
LGTM

Bidski (approved):

PR #49, opened 2021-05-22
Fix Jersey to Convert to Base Nodes

AlanaNoonan:


ysims (approved):
LGTM, everything works and the warning is gone :tada:

PR #50, opened 2021-05-23
Adds --clean option for configure

KipHamiltons:


scripts/configure.py
```diff
@@ -21,7 +29,14 @@ def register(command):
...
 
-def run(interactive, args, **kwargs):
+def run(purge_build_folder, interactive, args, **kwargs):
+
+    # Purge the build folder if we have passed the purge_build_folder argument
+    if purge_build_folder:
+        from shutil import rmtree
+
+        # We ignore errors so that it doesn't fail if `build` doesn't exist
+        rmtree(os.path.join(os.getcwd(), "build"), ignore_errors=True)
```
JosephusPaye, line 39: Should we clean this, instead of cwd/build? (from https://github.com/NUbots/NUbots/blob/master/tools/build.py#L22)

```
os.chdir(os.path.join(b.project_dir, "..", "build"))
```

TrentHouliston: Don't think you have the b script here do you? Or did you bring in the NUClear roles code.

JosephusPaye: Looks like we at least have [`b.project_dir` and `b.build_dir`](https://github.com/NUbots/NUWebots/blob/hamiltons/add_clean_option/scripts/b.py#L11-L13)

KipHamiltons: yea, I ported over some of the `b` stuff so we can configure/compile with the same commands. Not a full nuclear roles set up though

Bidski (dismissed):
LGTM

scripts/configure.py
```diff
@@ -21,14 +29,20 @@ def register(command):
...
+    # Purge the build folder if we have passed the purge_build_folder argument
+    if purge_build_folder:
+        from shutil import rmtree
 
-    # If interactive then run ccmake else just run cmake
-    os.chdir(os.path.join(b.project_dir, "build"))
+        # We ignore errors so that it doesn't fail if `build` doesn't exist
+        rmtree(b.build_dir, ignore_errors=True)
 
+    # Make sure we have a build directory the change into it
```
JosephusPaye, line 41:
```suggestion
    # Make sure we have a build directory, then change into it
```

JosephusPaye (changes requested):

JosephusPaye (approved):
LGTM

PR #54, opened 2021-05-24
Split the proto since the change got lost in nugus.proto merges

ysims:
Use the meshes that split the proto. Somewhere in the vision controller branch, a merge with main lost this stuff.


JosephusPaye (approved):
Looks fine in webots.

PR #55, opened 2021-05-24
Ignore files webots generates when converting a NUgus to base nodes

JosephusPaye:
Converting a NUgus to base nodes in webots generates `worlds/NUgus_textures` and `worlds/robot`, which we don't want to commit.


ysims (approved):
It works

PR #56, opened 2021-05-24
Fixes the textures inside mesh files for the RoboCup controller

ysims:
The textures seem to work fine in our worlds, but the ones in the separate mesh proto files can't find the textures when running the RoboCup controller because the `robots` folder doesn't exist. This changes them from being relative to the proto to relative to its own location (mesh folder).

### DIFF
--- a/protos/robot/NUgus.proto
+++ b/protos/robot/NUgus.proto
@@ -146,7 +146,7 @@ PROTO NUgus [
                 children [
                   DEF right_upper_arm NUgus_right_upper_armMesh {
                     robotTexture [
-                      %{='"./robot/NUgus_textures/NUgus_upper_arm_' .. color .. '.jpg"'}%
+                      %{='"../NUgus_textures/NUgus_upper_arm_' .. color .. '.jpg"'}%
                     ]
                   }
                   HingeJointWithBacklash {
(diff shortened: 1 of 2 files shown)


protos/robot/NUgus.proto
```diff
@@ -146,7 +146,7 @@ PROTO NUgus [
                 children [
                   DEF right_upper_arm NUgus_right_upper_armMesh {
                     robotTexture [
-                      %{='"./robot/NUgus_textures/NUgus_upper_arm_' .. color .. '.jpg"'}%
+                      %{='"../NUgus_textures/NUgus_upper_arm_' .. color .. '.jpg"'}%
```
JosephusPaye, line 149: Is the `NUgus_textures` folder autogenerated? Can't seem to find it in `protos/`.

ysims: https://github.com/NUbots/NUWebots/tree/main/protos/robot/NUgus_textures

JosephusPaye: Ah, so it's in `protos/robot`. It's weird how webots does the relative paths. 😅

JosephusPaye (approved):
LGTM

PR #57, opened 2021-05-25
Set the height based on model name

ysims:
Sets the ball height to its own height so it doesn't float, and adds the ball to the master teleport controller args. 
Set the lil Darwin to its own height as well because it was floating.


LachlanCourt (approved):
Does the job

PR #61, opened 2021-05-28
[WIP] Get Arms Out of Zombie Mode

KipHamiltons:
[Note: this PR is @moosahassanx's - Kip was just logged in to that lab computer when making the PR]
Arms of the NUGus robot need to be placed by its side rather than stick out.


wongjoel:
See also #63 for another approach to this.

PR #63, opened 2021-05-29
Vision collection controller to set shoulder motor positions randomly

wongjoel:
We need to have the robot's arms in various positions for the training data to be sufficiently diverse. This pull request adds the code required to move the shoulder motors in the arms.


wongjoel:
The current code doesn't actually work when observing in Webots. Waiting for Lachlan Court's fix for controller desync before investigating further.

wongjoel:
no longer needed due to #64

PR #67, opened 2021-06-01
Add in missing MX64-vel and MX106-vel in maxVelocity of motors

ysims:
Looks like they were missed when adding them in initially.


TheCoolKidsTable (approved):
LGTM!

PR #69, opened 2021-06-01
WEBOTS_HOME issue fix

KipHamiltons:
At the moment, if you run `./b configure` before running `source ~/.bashrc`, `WEBOTS_HOME` is set to `/usr/local/webots` in the CMakeCache. Since it's set in the cache, rerunning `./b configure` does not change its value, because of how we have it set up right now.
This results in a loop where the problem won't be fixed until the cache is reset and `./b configure` is re-ran with `WEBOTS_HOME` set correctly in the environment, or `./b configure -- -DWEBOTS_HOME=/path/to/webots` is ran.
We can avoid this loop by attempting to re-set `WEBOTS_HOME` if it happens to be the default value we give it.


CMurtagh-LGTM (approved):
LGTM

PR #70, opened 2021-06-02
Change the robot height in kid.wbt

ysims:
The height that the robot is dropped into the world is problematic for running our code on. It gives a negative z acceleration which makes the robot immediately get up. This height change works ok, with no get up triggered. This is also the height as defined in our team config for RoboCup.


LachlanCourt (approved):
Haven't tested but LGTM

PR #71, opened 2021-06-03
Move jerseys to the mesh folder

ysims:
This just tidies up our proto files a bit and puts the jerseys into the NUgus_mesh folder.


TheCoolKidsTable (approved):
LGTM 👍

PR #72, opened 2021-06-03
Adds center of mass to physics node of touch sensors

ysims:
We didn't have any center of mass so the automatic marking tool gave us warnings. This PR fixes this so we have a COM on the touch sensors. It's put to [0 0 0] because they are so small the actual position shouldn't be an issue.


TheCoolKidsTable (approved):
Lgtm!

PR #73, opened 2021-06-03
Reduce hip cylinder radius

TheCoolKidsTable:
This PR reduces the radius of the cylinder bounding box.

image


ysims (approved):
I think this is better. I don't think it'll let us do anything strange, and it'll allow us to use the scripts from the real robot without having to compensate for the large bounding box.

PR #75, opened 2021-06-03
Shift lower leg bounding box

Tom0Brien:
This PR raises the lower leg bounding box to hopefully prevent self collisions in the ankles. 

image


ysims:
I doubt merging in main will work (although it might tell you it does). If you want to get these changed in, I'd suggest making a new branch off the latest main and putting in your changes manually from here.
Although I think you wanted to redo the protos with that onshape stuff anyway?

ysims:
Is this still a problem @Tom0Brien?

PR #76, opened 2021-06-04
Change to using capsules for the bounding box for the cleats

ysims:
A requirement for playing with our robot in RoboCup 2021. Moves bounding objects on feet so they don't touch the cleats. Puts capsule bounding boxes on touch sensors that cover the cleats.


wongjoel (approved):
LGTM - tested and appears to work fine

PR #77, opened 2021-06-05
Updates the RoboCup field

ysims:
Looks like our field was a little out of date, and probably why I'm seeing differences in the get up between NUWebots and the RoboCup controller. This PR updates the field (and I'm getting similar stuff from the get ups with this vs the RoboCup controller).


LachlanCourt (approved):
Tested, LGTM

PR #83, opened 2021-08-03
Set up NUWebots with all required Webots files

ysims:
This is a change to the way we use the repos, and removes our need to use NUbots/webots. Instead we use the TC's webots fork as a subtree. 

- The TC fork of webots has been added as a subtree to the repo
- The `.gitmodules` file in the webots subtree has been moved out into the NUWebots directory
- Updates the subtree to use our GitHub mirrors for the Webots dependencies downloads
- Our `game.json` and `team_1.json` files from NUbots/webots have replaced the ones in the webots subtree
- Our `game.json` and `team_1.json` files have been updated for three robots instead of four or five
- Put the NUgus and the Wolfgang proto files into NUgus and Wolfgang folders - done so we can create a symlink to the folder in the Webots official RoboCup project.
- Updates the `.gitignore` to ignore the symlink that will be created in the RoboCup project proto directory. 

Will be merged (I hope 😅) at the same time as NUbots/NUbook#148, which will have updated instructions to work with what's in this PR.


robocup/game.json
```diff
@@ -0,0 +1,36 @@
+{
+  "type": "NORMAL",
+  "kickoff": "red",
+  "class": "KID",
+  "host": "127.0.1.1",
+  "minimum_real_time_factor": 0,
+  "side_left": "blue",
+  "supervisor": "test_supervisor",
```
LachlanCourt: This supervisor line breaks the build. Removing it fixes it, it doesn't exist in the standard build of webots so I'm not sure what it's purpose is

LachlanCourt (changes requested):
Chunky

I can't see on the diff bc it is too large but just making sure you've updated `README.md` as well
Also perhaps pin `NUWebots` repo now on our main github page instead of the `webots` fork

LachlanCourt (dismissed):
I am not sure why the build here failed, I think the test itself is misconfigured. I just ran the build from scratch for both webots and b script and both went without a hitch. b script has some readability warnings but no errors. This looks great, nice work!

KipHamiltons:
The build failing like that can be a few things. The `POST` size might be too large, or the network might have poor latency. If it's bad latency, trying again will probably fix it, but if it doesn't it'll need to be investigated, with the likely fix of changing the max size of the request.

This is based on [this SO link](https://stackoverflow.com/questions/15240815/git-fatal-the-remote-end-hung-up-unexpectedly)

LachlanCourt (approved):
LGTM

PR #86, opened 2021-12-13
Updates the postures.json so that the NUgus arms by its side in the upright posture

ysims:
Doesn't pass the auto model checking otherwise. Somehow this slipped through the cracks when updating this repo post-RoboCup.


CMurtagh-LGTM (approved):
lgtm

PR #87, opened 2022-01-10
Update Webots Subtree January 2022

ysims:
This pulls the Webots subtree in this repo to be up to date with the RoboCup Humanoid TC's Webots repo, as of 10/01/2022. 

There are a lot of changes, you may find it easier to check the commits (5ea3f37 and b01d1c9, the others will be a lot of changes).

It looks like Webots itself has had a major release and there should be changes from the TC with random background and ball textures, amongst other things.


ysims:
I have successfully compiled and ran this branch and it uses different backgrounds, balls, lighting.

JosephusPaye:
Looks like the file `webots/src/webots/vrml/WbProtoList.cpp` was auto formatted when you made the symlink change. Would it be possible to do that without formatting? So it's easier to see what actually changed. I think VSCode has a save without formatting feature.

JosephusPaye (approved):
LGTM if it works!

PR #90, opened 2022-02-04
Fix the NUgus robot's texture path

ysims:
It was relative to the protos folder rather than the proto itself. This caused issues in the RoboCup world because the location of the proto there was in the first level of the protos folder, whereas in our proto folder it's under a robot folder.

This PR changes it so the textures are relative to the main proto, `NUgus.proto`.


KipHamiltons (approved):
Seems ok

PR #92, opened 2022-03-20
Add a controller to interface with the Optimiser

wongjoel:
When running the optimiser (walk optimisation, script optimisation), we need additional messages beyond that in an ordinary game.

This currently fits additional messages in alongside the usual messages sent between NUbots and the robocup player controller.

I think long term, we probably want an "optimiser supervisor" that acts similarly to how the referee works in a game. (This way, we don't change anything about the comms with the robot, but we have a way to reset simulations and get exact locations of objects in the world)

Works with https://github.com/NUbots/NUbots/pull/540#issue-925411358


ysims (changes requested):
- The controller should be moved out from `webots/projects/samples/contests/robocup` and into `nuwebots` (which will soon just be the root directory when #99 is merged).
- There also should not be nugus textures in this PR.  I think once you move to the nuwebots directory, the gitignore will ignore it for you. 
- vscode settings should be reverted

@PiiDish

.vscode/c_cpp_properties.json
```diff
@@ -0,0 +1,17 @@
...
+            "defines": [],
+            "compilerPath": "/usr/bin/clang-10",
+            "cStandard": "c11",
+            "cppStandard": "c++14",
+            "intelliSenseMode": "linux-clang-x64",
+            "configurationProvider": "ms-vscode.makefile-tools"
+        }
+    ],
+    "version": 4
+}
```
ysims, line 17: remove

nuwebots/worlds/robocup_nuoptim.wbt
```diff
@@ -0,0 +1,66 @@
...
+    "1",
+    "127.0.1.1",
+    "10.1.0.102"
+    ]
+  }
+#1.571592653589793
+#Viewpoint {
+#  orientation 0.8938498316033444 -0.2697810913064417 -0.35812098698090605 1.4005817359063022
+#  position -8.658964597484593 -10.448717096347492 4.001015029323764
+#}
```
ysims, line 66: removed if commented out?

ysims (changes requested):
You shouldn't have any build folder or files generated by cmake here. You have a `controllers` folder as well in the root directory that you shouldn't have. You need to convert this from using a makefile to using our cmake set up. See the other controllers for info, they have their own `CMakeLists.txt` file that links everything they need. 
Remove added textures.

nuwebots/worlds/robocup_nuoptim.wbt
wongjoel: Yes, should remove before merge. The commented out bits are just saving useful constants.
1.571592653589793 is 90 degrees in radians
The Viewpoint defines the starting location of the camera, useful for getting a good starting camera angle on recordings. I don't remember which Viewpoint is which, but both aren't needed, so might as well remove the commented out one.

ysims:
@bencoady was there anything you wanted to change with the messages or can we get this merged now?

bencoady:
I think all the messages that I need are already there.

ysims (approved):
Should be good

ysims (approved):
lgtm

PR #93, opened 2022-04-02
Removes the webots folder from formatting

ysims:
`./b format` currently tries formatting all files in NUWebots, including those in `NUWebots/webots/`. This folder isn't our code, it's a submodule containing the Webots simulator code. We don't want to format files in this folder.

This PR updates the format script so it only formats our files and ignores the `webots` folder.


CMurtagh-LGTM (approved):
lgtm

PR #97, opened 2022-12-01
Add live Webots vision error testing (goal detection) 

BaileyJoe:
### Webots vision benchmarking for goal detection

- Adds message to support goal detection benchmarking using groundtruth data from webots.

https://github.com/NUbots/NUbots/pull/988/


ysims:
Need to merge main

PR #100, opened 2023-02-19
Updates rotation matrix to fix ground truth value

Jesse27:
Updates rotation matrix to fix ground truth value based on suggestion from @Tom0Brien

### DIFF
--- a/nuwebots/controllers/nugus_controller/nugus_controller.cpp
+++ b/nuwebots/controllers/nugus_controller/nugus_controller.cpp
@@ -759,10 +759,10 @@ public:
 
         // Get values from the robot model
         const double* rTXx = robot->getFromDef("BLUE_1")->getField("translation")->getSFVec3f();
-        const double* Rtx  = robot->getFromDef("BLUE_1")->getField("rotation")->getSFRotation();
+        const double* Rxt  = robot->getFromDef("BLUE_1")->getField("rotation")->getSFRotation();
 
         // Set values - need to convert angle axis to a rotation matrix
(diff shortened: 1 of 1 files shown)


nuwebots/controllers/nugus_controller/nugus_controller.cpp
```diff
@@ -272,7 +272,7 @@ class PlayerServer {
         const double* Rwx  = robot->getFromDef("BLUE_1")->getField("rotation")->getSFRotation();
 
         // Rotation is an angle axis so convert it to a rotation matrix
-        Hxw.linear() = Eigen::AngleAxisd(Rwx[3], Eigen::Vector3d(Rwx[0], Rwx[1], Rwx[2])).toRotationMatrix().inverse();
+        Hxw.linear() = Eigen::AngleAxisd(Rwx[3], Eigen::Vector3d(Rwx[0], Rwx[1], Rwx[2])).toRotationMatrix();
```
ysims, line 275: Should the name be swapped to Rxw if this is the case?
Lower down on L765, inverse is used. It's all the same maths, so does that need to be changed too?
@Tom0Brien

ysims: Have you tried with both changed? I guess here it doesn't do much since we  probably start pretty much on identity?

Tom0Brien: Yeah, should change both.

Tom0Brien: You're right, I didn't notice before because I was starting with robot upright, can start anywhere with any orientation if both changed.

Jesse27: Updated

ysims (approved):
sounds about right

PR #101, opened 2023-03-23
Adds a dockerfile for webots with nuwebots and robocup environments

JesseWilliamson:
## Requirements:
1. Host must be using X window server
2. host must be configured to allow local connections on the host machine. To do this, run `xhost +local:root`. To undo this run `xhost -local:root`.
3. For GPU acceleration the host computer must be using a NVIDIA graphics card and have nvidia-container-toolkit installed. `nvidia-container-toolkit` is available as an Ubuntu package and is on the AUR. Webots recommends using `nvidia-docker2` instead, but as far as I can tell this is deprecated.

## Notes:
1. The links on lines 27 and 60 reference files in this PR. For testing replace replace `main` with `williamson/webots-docker`.
2. Using docker for gui applications seems to have some fairly unpredictable problems. If the benefits of this docker image are ease-of-use and standardisation, these problems probably limit its usefulness significantly.
3. Most of this is from the cyberbotics [webots-docker ](https://github.com/cyberbotics/webots-docker)repo

## Running:
1. Build the image. There are arguments for building with different webots or ubuntu versions, but the defaults should be fine for nubots / robocup.
2. Find the ip of the container running nubots code on the host machine (I've just been using `ip addr` to get the address of the docker network device. `docker network inspect bridge` also works).
3. To run without gpu acceleration or ip whitelist and team name configured use `docker run -it -e DISPLAY -v /tmp/.X11-unix:/tmp/.X11-unix:rw nuwebots`
4. To run with GPU acceleration and configure with nubots team name and ip whitelist use `docker run --gpus=all -it -e DISPLAY --env TEAM_NAME=teams/nubots --env ROBOT_HOST=ip address from step 2 -v /tmp/.X11-unix:/tmp/.X11-unix:rw nuwebots`
5. Run webots in the container with world file as normal. Currently working with `webots ~/hlvs_webots/worlds/robocup.wbt `.

## Connecting with nubots robot code:
1. Get the ip from the webots docker container either by running hostname -i in the container or with `docker container inspect <container id> | grep IPAddress on` the host and enter in `./b edit config/Webots` as normal.

## Problems:
1. I'm not sure what the typical use case is for the nubots developed webots environment. I can run the world file as described in the [nubook page](https://nubook.nubots.net/guides/tools/webots-setup#runnypicalling-a-nubots-environment ). but I don't know how connecting to the robot would typically work, so I haven't been able to do it here.

### DIFF
--- a/webots-docker/Dockerfile
+++ b/webots-docker/Dockerfile
@@ -1,0 +1,63 @@
+ARG BASE_IMAGE=nvidia/cuda:11.8.0-base-ubuntu22.04
+FROM ${BASE_IMAGE} AS downloader
+
+# Determine Webots version to be used and set default argument
+ARG WEBOTS_VERSION=R2022b
+ARG WEBOTS_PACKAGE_PREFIX=
+
+# Disable dpkg/gdebi interactive dialogs
+ENV DEBIAN_FRONTEND=noninteractive
+
+RUN apt-get update && apt-get install --yes wget bzip2 && rm -rf /var/lib/apt/lists/ && \
+ wget https://github.com/cyberbotics/webots/releases/download/$WEBOTS_VERSION/webots-$WEBOTS_VERSION-x86-64$WEBOTS_PACKAGE_PREFIX.tar.bz2 && \
+ tar xjf webots-*.tar.bz2 && rm webots-*.tar.bz2
+
+FROM ${BASE_IMAGE}
+
+# Disable dpkg/gdebi interactive dialogs
+ENV DEBIAN_FRONTEND=noninteractive
+
+
+# Install Webots runtime dependencies
+RUN apt-get update && apt-get install --yes wget xvfb git python3-pip sudo && rm -rf /var/lib/apt/lists/ && \
+  wget https://raw.githubusercontent.com/cyberbotics/webots/master/scripts/install/linux_runtime_dependencies.sh && \
+  chmod +x linux_runtime_dependencies.sh && ./linux_runtime_dependencies.sh && rm ./linux_runtime_dependencies.sh
+
+# Install NUWebots dependencies
+RUN wget https://raw.githubusercontent.com/NUbots/NUWebots/main/webots-docker/dependencies.sh && \
+ chmod +x dependencies.sh && ./dependencies.sh && rm ./dependencies.sh && rm -rf /var/lib/apt/lists/
+
+# Install Webots
+WORKDIR /usr/local
+COPY --from=downloader /webots /usr/local/webots/
+ENV QTWEBENGINE_DISABLE_SANDBOX=1
+ENV WEBOTS_HOME /usr/local/webots
+ENV PATH /usr/local/webots:${PATH}
+
+# Enable OpenGL capabilities
+ENV NVIDIA_DRIVER_CAPABILITIES=graphics
+
+RUN useradd -ms /bin/bash webots && usermod -aG sudo webots && echo "webots ALL=(ALL:ALL) NOPASSWD: ALL" >> /etc/sudoers
+ENV HOME=/home/webots
+USER webots
+
+WORKDIR $HOME
+
+# Build the latest version of the RoboCup Humanoid TC fork of GameController
+RUN git clone https://github.com/RoboCup-Humanoid-TC/GameController ./GameController && cd GameController && ant
+
+# Build the robocup controllers
+RUN git clone https://github.com/RoboCup-Humanoid-TC/hlvs_webots.git ./hlvs_webots && \
+ pip3 install -r ./hlvs_webots/controllers/referee/requirements.txt && make -j$(nproc) -C ./hlvs_webots
+
+# Set environment variables for GameController
+ENV GAME_CONTROLLER_HOME=${HOME}/GameController JAVA_HOME=/usr/lib/jvm/java-11-openjdk-amd64
+
+# Install the NUbots-developed environment
+RUN git clone https://github.com/NUbots/NUWebots.git ./NUWebots && cd ./NUWebots && pip3 install -r ./requirements.txt && \
+ ./b configure -- -DENABLE_CLANG_TIDY=OFF && ./b build
+ 
+RUN wget https://raw.githubusercontent.com/NUbots/NUWebots/main/webots-docker/docker-config.py
+
+# Configure robocup game.json file based on environment variables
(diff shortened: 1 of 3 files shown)


ysims (approved):
Will need some documentation on NUbook for how to run this. Otherwise lgtm

webots-docker/Dockerfile
```diff
@@ -0,0 +1,66 @@
...
+FROM ${BASE_IMAGE} AS downloader
+
+# Determine Webots version to be used and set default argument
+ARG WEBOTS_VERSION=R2022b
+ARG WEBOTS_PACKAGE_PREFIX=
+
+# Disable dpkg/gdebi interactive dialogs
+ENV DEBIAN_FRONTEND=noninteractive
+
+RUN apt-get update && apt-get install --yes wget bzip2 && rm -rf /var/lib/apt/lists/ && \
```
KipHamiltons, line 11: Personally, I much prefer having each package in an install command in a dockerfile to be on its own line, as well as putting each command on its own line. Makes it much easier to read imo

webots-docker/Dockerfile
```diff
@@ -0,0 +1,66 @@
...
+
+# Install NUWebots dependencies
+RUN wget https://raw.githubusercontent.com/NUbots/NUWebots/main/webots-docker/dependencies.sh && \
+ chmod +x dependencies.sh && ./dependencies.sh && rm ./dependencies.sh && rm -rf /var/lib/apt/lists/
+
+# Install Webots
+WORKDIR /usr/local
+COPY --from=downloader /webots /usr/local/webots/
+ENV QTWEBENGINE_DISABLE_SANDBOX=1
+ENV WEBOTS_HOME /usr/local/webots
```
KipHamiltons, line 34: It still annoys me that webots needs this

webots-docker/Dockerfile
```diff
@@ -0,0 +1,66 @@
...
+ENV PATH /usr/local/webots:${PATH}
+
+# Enable OpenGL capabilities
+ENV NVIDIA_DRIVER_CAPABILITIES=graphics
+
+RUN useradd -ms /bin/bash webots && usermod -aG sudo webots && echo "webots ALL=(ALL:ALL) NOPASSWD: ALL" >> /etc/sudoers
+ENV HOME=/home/webots
```
KipHamiltons, line 41: Is this necessary?

webots-docker/Dockerfile
```diff
@@ -0,0 +1,66 @@
...
+USER webots
+
+WORKDIR $HOME
```
KipHamiltons, line 44: And this? I thought docker did this for you if you made your user with a home dir?

webots-docker/Dockerfile
```diff
@@ -0,0 +1,66 @@
...
+
+# Build the latest version of the RoboCup Humanoid TC fork of GameController
+RUN git clone https://github.com/RoboCup-Humanoid-TC/GameController ./GameController && cd GameController && ant
+
+# Build the robocup controllers
+RUN git clone https://github.com/RoboCup-Humanoid-TC/hlvs_webots.git ./hlvs_webots && \
+ pip3 install -r ./hlvs_webots/controllers/referee/requirements.txt && make -j$(nproc) -C ./hlvs_webots
```
KipHamiltons, line 51: I'm wondering - does $(nproc) affect caching if you ran it on a computer with a different amount of cores? Is it worth putting that in a script for that reason? (Presuming my suspicion is correct)

webots-docker/webots-config.py
```diff
@@ -0,0 +1,15 @@
+import json, os, sys
```
KipHamiltons, line 1: Is putting imports on separate lines good style in this codebase?

webots-docker/webots-config.py
```diff
@@ -0,0 +1,15 @@
...
+
+TEAM_NAME = os.environ.get('TEAM_NAME')
+ROBOT_HOSTS = os.environ.get('ROBOT_HOSTS')
+
+filename = sys.argv[1]
+
+with open(filename, 'r') as f:
+    game_config = json.load(f)
+
+if TEAM_NAME: game_config['red']['config'] = 'teams/' + TEAM_NAME + '.json'
```
KipHamiltons, line 11: Ifs on the same line as their bodies is definitely not my jam haha

webots-docker/entrypoint.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+python3 docker-config.py ${HOME}/hlvs_webots/controllers/referee/game.json
+rm docker-config.py
```
KipHamiltons, line 3: How come you're removing this file after you run it? It's unclear to me. What am I missing?

KipHamiltons (commented):
Quick and cheeky one off reviews

webots-docker/webots-config.py
TrentHouliston: I think this codebase needs to be blackened and isorted

webots-docker/Dockerfile
JesseWilliamson: That would make sense, but I haven't been able to find any solid info on it, so I'll need to test it. I don't understand why two computers would be building from the same cache though? Maybe I don't understand what you mean.

webots-docker/Dockerfile
JesseWilliamson: I tried removing this, but was getting errors. I don't think docker does it automatically. Either way, I've cleaned it up a bit now.

webots-docker/Dockerfile
JesseWilliamson: webots complains if it's run as root, so the user is needed, but I've cleaned up some of the other setup

webots-docker/Dockerfile
TrentHouliston: Doesn't change caching. All that would change it is if you changed the actual text here

webots-docker/Dockerfile
```diff
@@ -0,0 +1,99 @@
+ARG BASE_IMAGE=nvidia/cuda:11.8.0-base-ubuntu22.04
+FROM ${BASE_IMAGE} AS downloader
+
+# Determine Webots version to be used and set default argument
+ARG WEBOTS_VERSION=R2023b
```
ysims, line 5:
```suggestion
ARG WEBOTS_VERSION=R2022b
```

PR #109, opened 2023-09-28
Update optimisation for new robot model and better positioning

bencoady:


worlds/.nuoptimiser.jpg
ysims, line 1: Should this be here?

PR #115, opened 2024-11-29
Turn off clang tidy by default

ysims:
This has been sitting like this for ages and it causes compilation to fail for people setting it up for the first time. Might as well turn it off at this point.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -33,7 +33,7 @@ endif()
 set_project_warnings()
 
 # Default not to run the clang-tidy checks, default to whatever our CI_BUILD is
-option(ENABLE_CLANG_TIDY "Enable building with clang-tidy checks." ON)
+option(ENABLE_CLANG_TIDY "Enable building with clang-tidy checks." OFF)
 if(ENABLE_CLANG_TIDY OR CI_BUILD)
   find_package(PythonInterp 3 REQUIRED)
   set(CMAKE_CXX_CLANG_TIDY "${PYTHON_EXECUTABLE}" "${PROJECT_SOURCE_DIR}/scripts/clang-tidy.py"


CMakeLists.txt
```diff
@@ -33,7 +33,7 @@ endif()
 set_project_warnings()
 
 # Default not to run the clang-tidy checks, default to whatever our CI_BUILD is
-option(ENABLE_CLANG_TIDY "Enable building with clang-tidy checks." ON)
+option(ENABLE_CLANG_TIDY "Enable building with clang-tidy checks." OFF)
```
KipHamiltons, line 36: Can you fix what makes it fail the first time?

jpptm: Im not entirely sure as I am not well versed in clang-tidy but correct me if I'm wrong, I remember these things having a `warning-as-errors` option that would cause compilation to fail because clang-tidy is complaining about stuff. Maybe this is somewhere along those lines and just disabling clang-tidy will make it go away perhaps?

jpptm (approved):
lgtm, but see @KipHamiltons ' comment

CMakeLists.txt
jpptm: nvm `warning as errors` is off by default

ysims: @KipHamiltons  Can you elaborate, I'm not sure what you mean? Do you mean make it clang-tidy compliant?

KipHamiltons: Yes. Make it clang-tidy compliant. That would be swell :p 

Apologies for late reply haha.

ysims: That'll be a no hahaha, we've been sitting on a broken build for years, and new people will get stuck cause they don't know what the problem is. There are many issues because we're using cyberbotics code in our nugus controller, and it's just not a priority rn.

PR #116, opened 2025-04-27
Modify kid world to have an opponent and add 4v4 world

ysims:
This is for quick team play testing without the autoref.

### DIFF
--- a/worlds/1v1.wbt
+++ b/worlds/1v1.wbt
@@ -1,0 +1,88 @@
+#VRML_SIM R2022b utf8
+
+IMPORTABLE EXTERNPROTO "../protos/robocup_field/RoboCupBackground.proto"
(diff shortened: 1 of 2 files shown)


Tom0Brien (approved):
LGTM, can we also possibly change the name of the world? 

'kid.wbt' doesn't make a whole lot of sense tbh

ysims:
Change which? This PR renames `kid` to `1v1` as well.

Tom0Brien:
> Change which? This PR renames `kid` to `1v1` as well.

I just always had the thought that kid was confusing and I didn't see that you changed it hahaha.

Looks good!

ysims:
I agree it's always been a bit odd, I think it's left over from the kid vs adult distinction they had in the TC's repo.

PR #119, opened 2025-06-12
Fix multi robot  ground truth by using the robot's DEF name instead of hardcoding "BLUE_1"

ysims:
Turned out to not be a complicated fix, we were just hardcoding `BLUE_1` since we were only really testing one robot scenarios. But now for multi robot testing with ground truth, we need to dynamically get the DEF name.

### DIFF
--- a/controllers/nugus_controller/nugus_controller.cpp
+++ b/controllers/nugus_controller/nugus_controller.cpp
@@ -247,7 +247,8 @@ public:
                  int port,
                  int player_id,
                  int team,
-                 webots::Supervisor* robot)
+                 webots::Supervisor* robot, 
+                 std::string name)
         : allowed_hosts(allowed_hosts)
         , port(port)
         , player_id(player_id)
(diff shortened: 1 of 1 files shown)


controllers/nugus_controller/nugus_controller.cpp
```diff
@@ -979,8 +1000,9 @@ int main(int argc, char* argv[]) {
     const std::string name    = robot->getName();
     const int player_id       = std::stoi(name.substr(name.find_last_of(' ') + 1));
     const int player_team     = name[0] == 'r' ? RED : BLUE;
+    std::string player_name = std::string(name[0] == 'r' ? "RED" : "BLUE") + "_" + std::to_string(player_id);
```
jpptm, line 1003:
```suggestion
    std::string player_name = std::string(player_team) + "_" + std::to_string(player_id);
```

jpptm (approved):

PR #121, opened 2025-07-08
Herfel/ground truth obstacles

A-Herfel:


ysims (approved):
For future, might be good to generalise this a bit

